PR #4, opened 2015-11-11
[android] updated notification frostwire icon

marcelinkaaa:
So the reason that the FrostWire notification icon looked like a nontransparent white ball is that starting in Android Lollipop all the icons in status bar are required to be white (although at times apps will hack this to use their colored version). And we were using our colored version without any modifications so that android was paining it white. 

I think Alden did that on purpose as it was causing some problems on lollipop, however now everything seems to be working fine after the recent updates to android app. Tested it on lollipop, marshmallow and api 16 emulator.

### DIFF
--- a/android/apollo/src/com/andrew/apollo/NotificationHelper.java
+++ b/android/apollo/src/com/andrew/apollo/NotificationHelper.java
@@ -118,7 +118,7 @@ public class NotificationHelper {
     }
 
     private int getNotificationIcon() {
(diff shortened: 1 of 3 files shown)


gubatron:
wohoo, so cool to see this being taken off my hands.

I'm going to test this branch on 4.4. and 5.0 and see what happens.

marcelinkaaa:
awesome, thank you.

gubatron:
perfect on Lollypop

gubatron:
perfect on 4.4

gubatron:
Kitkat I think it was called

gubatron:
thanks-fist-bump

PR #7, opened 2015-11-15
[desktop] upd translations

X-NicON:
fix gettextExtract in gradle file
sync translation with src
update Ru translation


gubatron:
great catch!, thanks
its_all_you

PR #9, opened 2015-11-16
[android] Sleep 1 more second, make sure BTEngine is stopped.

gubatron:
@aldenml I couldn't replicate the issue with my phone. But I suppose when it happens it must be some sort of syncrhonization issue with the Service's onDestroy method.

My guess is that the process is killed before the Bittorrent session can finish it's `abort()` procedure.

Therefore I propose this patch to have that last thread sleep for one more second.

If not, maybe a better patch would be to somehow get a valid confirmation from the bittorrent session that it's actually been aborted all the way, and then we can continue to shutdown.

### DIFF
--- a/android/src/com/frostwire/android/gui/services/EngineService.java
+++ b/android/src/com/frostwire/android/gui/services/EngineService.java
@@ -116,7 +116,7 @@ public class EngineService extends Service implements IEngineService {
             @Override
             public void run() {
                 try {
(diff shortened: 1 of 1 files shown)


aldenml:
I don't think it's a sync issue but us doing something not supposed to do (killing the process). I will merge this since it probably alleviates the issue.

PR #22, opened 2015-12-08
Wizard page update

marcelinkaaa:
Still needs work. Current Storage Location is not functional yet, just a dummy text. 
Also the "Seed only on Wifi" should be fully unchecked while 'Seed Finished Torrents" is also unchecked. 
As of now it still looks like seed only on wifi is preselected for the user and a default action.


gubatron:
Thank you

Now, a shameless plug.

Support FrostWire, get [FrostWire stickers](http://shop.frostwire.com)
cvzh96zwcaauid7

PR #35, opened 2016-01-03
[desktop] spanish translation updates.

alejandroarturom:


desktop/lib/messagebundles/es.po
```diff
@@ -2441,11 +2433,11 @@ msgstr ""
 
 #: /frostwire/desktop/src/com/limegroup/gnutella/gui/GUIMediator.java:113
 msgid "&Search"
-msgstr "Bu&scar"
+msgstr "&Buscar"
```
gubatron: I'm not too sure of this, it might work though...
not-impressed

gubatron:
will merge for now, those shortcut accelerators should work, but we need to test the UI in spanish to make sure. They'll be a lot more comfortable to use for sure if they end up like this.

Issue #51, opened 2016-01-05
[android] When opening My Music for the first time it should show "Last Added" tab, after that the application should remember


gubatron:
If you've never played a song and the first thing you did was to download a song, the first time you open the Music Player you will see an empty list because you've never played anything.

At this moment you should be displayed the "Last Added" tab.

Then, from there on, we should keep track of what was the last tab touched on an internal setting variable (See `PreferenceUtils.java` which is where it seems apollo keeps its preferences, there seems to be something related `setStartPage()`).

Classes to look at inside `android/apollo`:
`ProfileActivity.java`, `PagerAdapter.java`, `MusicBrowserPhoneFragment.java`

CC: @votaguz @alejandroarturom (@arubinofaux wanna do some android?)

$15 bounty for merged pull request to be paid in Bitcoin at the exchange rate of the day on Bitven.com

marcelinkaaa:
@gubatron so its not that you want the "Last Added" tab to be always first when My Music is opened for the first time, right? What you want is for it to be first when you just go to My Music without playing or downloading anything or just downloading something. If you played some of your content and then you open My Music for the first time you would expect to see Recent tab? 

Just wanted to clarify as I already have it set up so it always opens to the Last Added tab and later on remember which tab you were last. But I dont really know how to do the other part. Let me know.

gubatron:
>so its not that you want the "Last Added" tab to be always first when My Music is opened for the first time, right
yes, just the first time.

once you start playing stuff you should see Recent, or whatever last tab you left selected before you leave My Music.

gubatron:
@marcelinkaaa check inside the apollo folder `PreferenceUtils.java`

marcelinkaaa:
I got it already and changed it to page 0 - which is Last Added as a Default Page. Than the code remembers what was your last tab. This is done. 

However how about this scenario - you first get the app installed and go to my files, for ex, playing one of  your current songs you already had - thus opening the player, after that you go back to navigation (without being in My Music) and select My Music for the first time - should it still show you recent? Or no matter what you did before the first page you see should be Last Added?

PR #52, opened 2016-01-05
[android] changed the default page to Last Added in My Music

marcelinkaaa:
This works just fine if you want the Last Added to be the first tab when you open My Music for the first time. See my further comment on #51 if you think the Recent tab should be also a default page in My Music depending if you played one of your songs first without downloading.


gubatron:
thanks-playa

PR #60, opened 2016-01-06
Android icon update

marcelinkaaa:
per discussion with @aldenml removed xxxhdpi icons from the drawer, added mipmap folders for updated launcher icon


gubatron:
no-testing-required-it-will-work

Issue #76, opened 2016-01-16
Update Copyright Dates


johnathanmartin:
Copyright dates should be changed from 2015 to 2016

gubatron:
The copyright dates on file headers get updated as they are edited.
If the file hasn't been touched we leave the file copyright as is.
We learned this practice from the Bitcoin-Core project.

PR #77, opened 2016-01-16
Issue #76

johnathanmartin:
issue #76


johnathanmartin:
Update expiration dates

gubatron:
The copyright dates on file headers get updated as they are edited.
If the file hasn't been touched we leave the file copyright as is.
We learned this practice from the Bitcoin-Core project.

gubatron:
The copyright dates on file headers get updated as they are edited.
If the file hasn't been touched we leave the file copyright as is.
We learned this practice from the Bitcoin-Core project.

PR #86, opened 2016-01-31
[desktop] Italian Translation Update 1.

alejandroarturom:
I edit some lines because there were erroneous translations. Again, if there is any italian native speaker reading this, please check the translation :).


gubatron:
thanks-playa

PR #100, opened 2016-03-08
ExpressVPN update

marcelinkaaa:
@gubatron here are the initial changes for the suggested improvements to the VPN-detail page. Links also changed. And per suggestion both - learn more and get VPN - are directed to the provider's website. 
Have @kad3mlia show you the suggested changes in the email. 
The only thing that is not yet done are the notifications on the transfer screen which I am not sure how to do, if you have some pointers I will be more than happy to try.


android/res/values/strings.xml
```diff
@@ -21,7 +21,16 @@
 <resources>
     <!-- Important: Android is buggy with HTML, do not put HTML tags right next to text, leave a space, as it seems the text wrapping logic is considering HTML as visible text and you get premature wrapping. -->
     <string name="your_privacy_is_a_fundamental_right_html">Your privacy is a fundamental human right.</string>
-    <string name="you_dont_need_a_vpn_to_use_frostwire_bullet_html"><![CDATA[<p> * Although highly recommended, <b>you do not need a VPN to use FrostWire. </b></p>]]></string>
+    <string name="you_dont_need_a_vpn_to_use_frostwire_bullet_html"><![CDATA[* Although recommended for your downloads, <b>you do not need a VPN to use FrostWire. </b>]]></string>
```
aldenml: I never liked the idea of having the CDATA trick, but in any case, this seems to me like too much. String resources are better suited for small messages, for full text, we should use raw content.

android/res/values/styles.xml
```diff
@@ -370,12 +370,15 @@
 
     <style name="GetVPNButton">
         <item name="android:background">@drawable/vpn_button_background_selector</item>
-        <item name="android:textColor">#3c4310</item>
+        <item name="android:textColor">@color/white</item>
```
aldenml: Is this the apollo white? I think android color should be android:color/white

aldenml:
Minor comment: the naming of variables in the source code part is a little odd and off of our style, but it takes some time to develop this, specially if we don't have our guidelines written.

android/res/values/styles.xml
marcelinkaaa: Thanks. Yes it is, autocomplete. Will change for the basic_white from FW colors.

android/res/values/strings.xml
marcelinkaaa: raw content? What is the case against CDATA? I didn't really read anything really bad about it?

aldenml: Ideally, you should not have formatting inside the string resources, strings res are more for data (minor formatting like \n is aceptable), but if you look at `VPN_feature_list_html`, that's is a higher level of formatting.

The use of CDATA (together with the comments by angel) is a strong indication that this is not the best way. If you use raw resources, that's a clear way to say that you need to put a lot of stuff that does not fit in a short single line <string> concept.

Can you explain to me the idea of `protected_connections_visibility_bullet_html`. If you are using a VPN, are you still recommending another one?

android/src/com/frostwire/android/core/Constants.java
```diff
@@ -180,8 +180,8 @@ private Constants() {
     public static final String SOCIAL_URL_TWITTER_PAGE = "https://twitter.com/frostwire";
     public static final String SOCIAL_URL_REDDIT_PAGE = "https://reddit.com/r/frostwire";
 
-    public static final String VPN_LEARN_MORE_URL = "http://www.frostwire.com/vpn";
+    public static final String VPN_LEARN_MORE_URL = "https://www.expressvpn.com/internet-security/private-vpn/buy?a_aid=frostwire";
```
gubatron: nope. we can change that on the website. let's not change this url please.

gubatron: if you need another URL to directly link to expressvpn, then use another variable.

android/src/com/frostwire/android/gui/activities/VPNStatusDetailActivity.java
```diff
@@ -58,35 +60,78 @@ protected void initComponents(Bundle savedInstanceState) {
         
         final ImageView headerIcon = findView(R.id.view_vpn_status_header_icon);
         final TextView headerTitle = findView(R.id.view_vpn_status_header_title);
-        final Button button = findView(R.id.view_vpn_status_get_vpn_button);
+        final TextView unProtectedTitle = findView(R.id.view_vpn_status_header_un_protected);
```
gubatron: unprotected is a single word, no need to camelcase the P

android/res/values/strings.xml
```diff
@@ -21,7 +21,16 @@
...
+    <string name="unprotected_connections_visibility_bullet_html"><![CDATA[ An unprotected connection means everything you do online (including <b>searches, messaging, and downloads</b>) may be monitored by anyone on your network and even by your Internet service provider.<br><br>We recommend using ExpressVPN to encrypt your connection and <b>safeguard your privacy</b>.]]></string>
+    <string name="protected_connections_visibility_bullet_html"><![CDATA[ You\'re connected to the Internet using a VPN. <br><br> If you are not already, we recommend using <b>ExpressVPN</b> to encrypt your connection and <b>safeguard your privacy</b>. ]]></string>
+    <string name="VPN_feature_list_html"><![CDATA[
```
gubatron: yeah, no way.

You want each bullet as an indivirual string @marcelinkaaa

android/res/layout/view_vpn_status_detail.xml
```diff
@@ -48,49 +48,130 @@
...
+        <TextView
+            android:id="@+id/view_vpn_status_header_un_protected"
+            style="@style/VPNTitle"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:gravity="center_horizontal" />
+
+        <!-- VPN description text -->
+        <TextView
+            android:id="@+id/VPNText"
```
gubatron: rename to `view_vpn_status_VPNText`

android/res/layout/view_vpn_status_detail.xml
```diff
@@ -48,49 +48,130 @@
...
+            style="@style/VPNSmallText"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:gravity="center_horizontal" />
 
-        <!-- explanation about what is open -->
         <TextView
-            android:id="@+id/VPNText"
-            style="@style/VPNText"
+            android:id="@+id/view_vpn_price"
```
gubatron: rename to `view_vpn_status_vpn_price`

android/res/layout/view_vpn_status_detail.xml
```diff
@@ -48,49 +48,130 @@
...
+            android:id="@+id/view_vpn_status_learn_more_button"
+            style="@style/GetVPNButton"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:layout_marginTop="30dp" />
+
+
+        <ImageView
+            android:id="@+id/imageView"
```
gubatron: does this need an id? if so, please make the name self explanatory, perhaps:
`rename to `view_vpn_status_divider_image_view`. If you are not using this view in your java code, I believe you can just get rid of the id.

android/res/layout/view_vpn_status_detail.xml
```diff
@@ -48,49 +48,130 @@
...
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:layout_marginBottom="15dp"
+            android:layout_marginTop="5dp"
+            android:gravity="center_horizontal"
+            android:orientation="horizontal">
+
+            <TextView
+                android:id="@+id/VPN_what_do_you_get"
```
gubatron: rename to `view_vpn_status_what_do_you_get`

android/res/layout/view_vpn_status_detail.xml
```diff
@@ -48,49 +48,130 @@
...
+                style="@style/VPNTitleSecond"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:paddingRight="5dp"
+                android:text="@string/VPN_what_you_get_with" />
+
+            <TextView
+                android:id="@+id/VPN_client"
```
gubatron: rename to `view_vpn_status_client_textview`

android/res/layout/view_vpn_status_detail.xml
```diff
@@ -48,49 +48,130 @@
...
+            style="@style/VPNText.Regular"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:gravity="center_horizontal"
+            android:text="@string/VPN_feature_list_html" />
 
         <!-- Bullet - VPN not required for FrostWire but highly recommended -->
-        <com.frostwire.android.gui.views.HTMLTextView
+        <TextView
+            android:id="@+id/vpn_bullet"
```
gubatron: rename to `view_vpn_status_bullet_textview`

android/res/layout/view_vpn_status_detail.xml
```diff
@@ -48,49 +48,130 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:layout_marginBottom="10dp"
+            android:layout_marginTop="15dp"
+            android:src="@drawable/express_vpn" />
+
+        <!-- GetVPN bottom button -->
+        <Button
+            android:id="@+id/button_VPN"
```
gubatron: rename to `view_vpn_status_get_vpn_button`

gubatron: I'm anal about naming because otherwise it's really easy to have a mess and confuse ids and have null pointers.
Let's stick to our conventions they've worked out great over the years.

gubatron: by having the name of the xml file as a prefix, with intellij or android studio, it's pretty awesome how you can have all the resources of a layout visible at once.

android/src/com/frostwire/android/core/Constants.java
gubatron: I'm thinking maybe if you want to have a direct link to a partner, and we want to change these URLs, it's better to do so remotely, perhaps this link could be instead something like `http://www.frostwire.com/vpnpartner1`, `http://www.frostwire.com/vpnpartner2`...

and then we point to the desired locations on the server.

android/res/values/strings.xml
marcelinkaaa: Yes. Per suggestions on improving conversion of the VPN purchases - the company recommended to include suggestion for their service when a user is already using a VPN provider. 

If I would like to keep the formatting than, what is the best way to approach it?

android/src/com/frostwire/android/gui/activities/VPNStatusDetailActivity.java
marcelinkaaa: sorry will rename. Was using the TextView for displaying both the protected and unprotected text based on your connection. But will change it for better.

android/res/values/strings.xml
marcelinkaaa: will do in all the single strings.

android/res/values/strings.xml
aldenml: I can live for a while with `unprotected_connections_visibility_bullet_html` and `protected_connections_visibility_bullet_html` until I figure a better solution.

gubatron:
12325550_10154043413017863_1156724544_o

@marcelinkaaa now we just gotta add some logic to this, when to show it, when to hide it.

If that's too much padding there, I think you can adjust it easily. Take a look at the last commit so you understand how it works now.

I suppose I read the same thing as you, but you understood it otherwise.

When I looked on how to float elements on top, I found that this is what relative layout does, if you don't specify how elements relate within the layout, they are drawn on top of each other.

So I figured, what needed to be on the RelativeLayout, wasn't the TextView, it was both the Expandable List View, and then we wanted to put the text view on top of it, aligned to the bottom.

I took out the text view you had on that separate relative layout which only had the textview, and moved it over here, and it worked, yay.

marcelinkaaa:
Great, looked over your commit. Thanks and got it :) 
As to the logic. Based on the given recommendation, which I agree with, we should show 'Your connection might not be private' when VPN is not detected and briefly show 'You're connected to VPN' when they are. I would have it show the first time user come to the transfer screen for about 5 seconds. Maybe show it for the first 2-3 sessions and than not show it again, if possible?

gubatron:
hmm, show for 2-3 sessions, interesting.

We were talking over here about showing it for 10 seconds, every new session, when the user first opens the Transfer's fragment and VPN is off.

I'll start coding away and see how easy this will be (like keeping track of when the Fragment screen was shown the first time), or if there's some simpler more stupid approach that works just as well.

I think I wouldn't do it only for a few sessions, because I believe it's important that users know when their privacy is at stake.

marcelinkaaa:
So having it display for few sessions is used to bring users' attention to the icon which is always there to show people if they are connected or not. When you get use to the presence of the icon and fully understand what it means having a 'secondary' notification is not necessary but could be a bit too much if it always displays for 10 sec, for ex. 
We could change the number of sessions but I would not leave it there forever, or maybe remind a user about the VPN every few sessions after that?

gubatron:
good point. Will do it for maybe 5 sessions then.
What about making it dissapear when the fragment is hidden?

Also, if you touch it, do you want it to go away, or do you want it to go away and take the user to the VPN screen?

marcelinkaaa:
Yes, that sounds like a good start. If we see it not working in some way and get good feedback we could adjust accordingly. 
Personally I would make the user go to the VPN page on the click of the toast, unless you want to set the time of the toast display for something longer - than I would give a user an ability to dismiss it by tapping on it. I would say 8-10 sec and up. Of course this is not backed by data just what I think, up for discussion.

gubatron:
merging ciocia malina.

PR #119, opened 2016-04-25
[android] Adding PAD File.

alejandroarturom:
I am adding a PAD File because some App Directories ask for it in order to accept the app in their site.
--
Portable Application Description (PAD) is a machine-readable document format designed by the Association of Shareware Professionals.
It allows authors to provide product descriptions and specifications to online sources in a standard way, using a standard data format, a simplified subset of XML, that will allow webmasters and program librarians to automate program listings. PAD saves time for both authors and webmasters.
Each field in the specification has a regular expression (regex) associated with it. The regex acts as a constraint on the field: if the regex matches, the field value is legal and if it fails to match, the field and the PAD file as a whole are out of spec. Only files where all fields in the file pass validation are properly called PAD files.
The simplifications in PAD over XML are primarily PAD does not use name/value pairs in tags. All tags are attribute-free. This is less expressive than XML but easier to parse. The official PAD spec uses unique tags. To extract the fields in the official spec, it is not necessary to descend through the tag path. However, if multiple languages are represented in a single PAD file, then correct parsing does require descending though the tag path because leaf tags are duplicated for each language supported.


gubatron:
looks like the file you were adding was actually your work on portuguese, if you lose the file, the changes are here on this patch (just in case)

PR #129, opened 2016-05-17
List view design refresh

marcelinkaaa:
search_result_refresh_2
search_results_refresh

Let me know if you would consider merging these slight changes in the look of search results and browse peer list. Attaching images. I think its a lot cleaner to read, also changed colors are more consistent with the rest of the app.


aldenml:
I kind of like the lighter blue and the highlighted blue for the size. Why is this better to read

marcelinkaaa:
Well we are using one kind of blue for the title, another blue for the source, and another blue for the size - not great to highlight what is the most important - except for the size as it is the first thing that my eyes go to, and in my opinion it should be the title. The blue used for the size also looked almost too bright in one of my devices so I can just imagine the variations on different screens. Also this new way makes the the source more prominent as a link (which is still a good way to go to the source to check comments or the health of the torrent). 

Another thing the colors are now consistent with list view in other places in the app like the apollo player.

aldenml:
Good points. I don't have strong opinions about this.

PR #130, opened 2016-05-17
Updating branch Espresso-2 - Adding test 'Search'

muckachina:
Search test check we are getting at least one result under each label.


gubatron:
ha, you closed it :)

Yeah, we wouldn't merge pull requests like these.

Make sure the pull request contains only the commits you made.

You can start a new branch that stems off of the most recent point on master, apply your changes and then send the pull request.

If in the meantime master keeps moving, then you should rebase your changes, so they appear at the top.

Issue #116, opened 2016-04-12
[android] Repositioning tracks in Playlists does not work ($15 bounty)


kad3mlia:
When you open any playlist and try to drag and drop tracks into different positions, the new position either does not take, on Sony Xperia Z5/ Android 6 it just hangs mid new position (sort of on top of two other tracks). User said it 'glitches'

gubatron:
@votaguz can you help me with this one? [$15 bounty]

votaguz:
Yes sir

gubatron:
awesome!

gubatron:
Start by taking a look at the PlaylistFragment and its adapter.
I know it has a class somewhere in there that takes care of the drag and drop and the reordering, might be something stupid related to the number of items it's expecting (offset)

votaguz:
Still working on grid thing. But i will take a look.

gubatron:
@votaguz what is the update on this?

votaguz:
Still trying to figure out how Apollo inflate the container that tracks the movement because seems that is inflated with the incorrect dimensions, that is why when you try to putting again the list doesn't know how to reorganize the elements. I apologize for the delay.

grzesiekrzaca:
https://github.com/frostwire/frostwire/pull/136

gubatron:
Fixed by @grzesiekrzaca

PR #136, opened 2016-05-20
[android] issue #116 repositioning songs in playlist

grzesiekrzaca:
Everything was "ok" in the View. The problem lay with no listener that would change the adapter (and data in system). The problem with the adapter is that it uses a copy of the data other than the one in ArryAdaper (why? 0.o) so that copy also has to be updated.

### DIFF
--- a/android/apollo/src/com/andrew/apollo/adapters/ProfileSongAdapter.java
+++ b/android/apollo/src/com/andrew/apollo/adapters/ProfileSongAdapter.java
@@ -60,6 +60,8 @@ public class ProfileSongAdapter extends ApolloFragmentAdapter<Song> {
      */
     private final LayoutInflater mInflater;
 
+    private final Object mDataListLock = new Object();
+
     /**
      * Fake header
      */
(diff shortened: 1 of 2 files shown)


gubatron:
just saw this, will test and review your branch. Thank you for submitting this.

gubatron:
note: next time please submit your pull request on a separate branch.

gubatron:
awesome, it works.
@aldenml this LGTM, please take a look before we merge.

In the meantime @grzesiekrzaca please send me your bitcoin payment address to send you the bounty payment to gubatron@gmail.com

Then let's try another ticket to see how you work with more complex/larger problems.
So far 👍 👍

aldenml:
Hi @grzesiekrzaca, this part of the code is not the one I'm most familiar with but it LGTM. One question though: Why the `mDataListLock` and the `synchronized` blocks?

grzesiekrzaca:
So that any concurrency problems don't arise from the fact that there are multiple resources that represent the same thing in this class (due to inheritance from ArrayAdaper). Though I see I made a mistake and got that super call outside of critical section. (Which is also not perfect as in this case we notify observers of the list inside a critical section, though in normal application this should not be a problem).
PR to fix this (update pr from a new branch - i hope this is it): #139
Also sorry for VC mess. I'm used to bitbucket and feature branches setup.

Issue #138, opened 2016-05-21
[android] Optimize UI refresh in transfers


aldenml:
Currently, there is a `TimerService.subscribe` to refresh the transfer list every 2 seconds, but this mechanism does not take in consideration if the view is actually visible or not. Ideally a general solution tied to the `TimerService` should free any UI component subscription of this problem.

aldenml:
Hi @grzesiekrzaca, may be you want to take a look at this

grzesiekrzaca:
In my opinion, the caller should clear it's subscription to anything it doesn't need (he knows best what he needs). In the case of the fragment, it should be in the `onStopped()` method (no longer visible). The problem here is that the notification that shows the download/upload also is updated in the fragments `onTime()` method (which is not ideal due to SRP). I propose adding normal removal of subscription to the fragment and moving the notification update to separate class. 
Clarification:
Due to weird way the fragments are handled here this approach doesn't work (just doing everything in onStopped/onResumed) so I added a check for visibility inside onTime(). If it doesn't pass then the subscription is cancelled. (Also not ideal but will end in at most one redundant call of the timer (without UI logic)) #140

aldenml:
The corresponding call to `onResume` is `onPause`. The way the PR is coded right now, it's possible you  end up creating adding new subscriptions without removing the old ones.

The idea of having a solution more inside the "Timer" framework is that all this UI components potentially have the same problem of "don't refresh if not necessary". This `TimerService` is just an UI timer running in the main looper for simple UI refresh, why not put the solution here (not necessarily inside the class `TimerService`?

I do agree with the idea of moving out the task of updating the permanent status notification from the transfer fragment.

Ok, I think it's a good idea that we start with the permanent status "cleanup" that should be done in the right place (and in a separated PR).

grzesiekrzaca:
Hmmm add `isSubscriptionStillValid()` to the interface and make the Timer ask the subscribed UI element?

aldenml:
What if instead of (re)subscribing, we just let the `TimerSubscription` decide if effectively call the `onTime`. For that, I would ask if the observer is (or is in) an `Abstract(Activity|Fragment)`. We can add one more state to them if necessary.

grzesiekrzaca:
[...]

PR #140, opened 2016-05-23
 Fewer timer events for transfers fragment ui

grzesiekrzaca:
move notification handling to another class  
check in onTime if the subscription is valid.

### DIFF
--- a/android/src/com/frostwire/android/gui/fragments/TransfersFragment.java
+++ b/android/src/com/frostwire/android/gui/fragments/TransfersFragment.java
@@ -121,6 +121,11 @@ public class TransfersFragment extends AbstractFragment implements TimerObserver
     public void onResume() {
         super.onResume();
         initStorageRelatedRichNotifications(getView());
+    }
+
+    @Override
+    public void onStart() {
+        super.onStart();
         subscription = TimerService.subscribe(this, UI_UPDATE_INTERVAL_IN_SECS);
         onTime();
     }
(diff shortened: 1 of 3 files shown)


aldenml:
See my comment in https://github.com/frostwire/frostwire/issues/138

android/src/com/frostwire/android/gui/fragments/TransfersFragment.java
```diff
@@ -119,21 +118,21 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
+        subscription = TimerService.subscribe(this, UI_UPDATE_INTERVAL_IN_SECS);
         onTime();
     }
 
     @Override
-    public void onDestroyView() {
-        super.onDestroyView();
+    public void onStop() {
+        super.onStop();
         subscription.unsubscribe();
```
gubatron: @aldenml have we checked if it actually unsubscribes when the fragment is invisible (I assume this un-subscription was here before), from the documentation I understand it only stops because the activity stopped.

>onStop() fragment is no longer visible to the user either because its activity is being stopped or a fragment operation is modifying it in the activity.

android/src/com/frostwire/android/gui/fragments/TransfersFragment.java
```diff
@@ -153,6 +152,10 @@ public void onPause() {
 
     @Override
     public void onTime() {
+        if(!this.isVisible()){
```
gubatron: I guess this would be do the trick on the next refresh then

gubatron:
what happened with this discussion, did @grzesiekrzaca arrive to the simpler fix?

aldenml:
I didn't receive a notification for the last commit in https://github.com/frostwire/frostwire/pull/144. Looking at it.

PR #141, opened 2016-05-23
Android permanent notification deamon

grzesiekrzaca:

PR #144, opened 2016-05-24
Android permanent notification deamon

grzesiekrzaca:
Ups
continuation of #141 
sorry for making all that mess

Yes, this should reside inside service, but I think NUD has its place. EngineService is almost 400 lines. I'm even thinking about removing the whole updatePermanentStatusNotification(...) from the EngineService into the demon but IDK if there's a reason that it's there (nothing seems to call it). Any thoughts?

### DIFF
--- a/android/src/com/frostwire/android/gui/NotificationUpdateDemon.java
+++ b/android/src/com/frostwire/android/gui/NotificationUpdateDemon.java
@@ -1,0 +1,137 @@
+package com.frostwire.android.gui;
+
+import android.app.Notification;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+import android.support.v4.app.NotificationCompat;
+import android.widget.RemoteViews;
+
+import com.frostwire.android.R;
+import com.frostwire.android.core.ConfigurationManager;
+import com.frostwire.android.core.Constants;
+import com.frostwire.android.gui.activities.MainActivity;
+import com.frostwire.android.gui.services.EngineService;
+import com.frostwire.android.gui.transfers.TransferManager;
+import com.frostwire.android.gui.util.UIUtils;
+import com.frostwire.android.gui.views.TimerObserver;
+import com.frostwire.android.gui.views.TimerService;
+import com.frostwire.android.gui.views.TimerSubscription;
+import com.frostwire.logging.Logger;
+
+public class NotificationUpdateDemon implements TimerObserver{
+
+    private static final Logger LOG = Logger.getLogger(NotificationUpdateDemon.class);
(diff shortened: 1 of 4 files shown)


aldenml:
Good, and yes, let's remove `updatePermanentStatusNotification` from EngineService.

If you think NUD needs an ApplicationContext (and I think it does) let's put that type in the constructor and inner field.

grzesiekrzaca:
I just saw that SettingsActivity just cancels the notification instead of communicating with the service and turning it on/off. Should I add that functionality here or create a(n issue and a) new PR for only that?

Also please check slack.

aldenml:
Let's review the SettingsActivity later, there is plenty of room for refactor after this.

android/src/com/frostwire/android/gui/NotificationUpdateDemon.java
```diff
@@ -0,0 +1,63 @@
...
+    private TimerSubscription subscription;
+
+    public NotificationUpdateDemon(Context parentContext) {
+        LOG.debug("Creating permanent notification demon");
+        this.parentContext = parentContext;
+    }
+
+    public void start(){
+        LOG.debug("Starting permanent notification demon");
+        if(subscription!=null){
```
aldenml: Format this with spaces (and review the rest of the code for standard formatting)

android/src/com/frostwire/android/gui/NotificationUpdateDemon.java
```diff
@@ -0,0 +1,63 @@
...
+    private void updatePermanentStatusNotification() {
+        //  format strings
+        String sDown = UIUtils.rate2speed(TransferManager.instance().getDownloadsBandwidth() / 1024);
+        String sUp = UIUtils.rate2speed(TransferManager.instance().getUploadsBandwidth() / 1024);
+
+        // number of uploads (seeding) and downloads
+        int downloads = TransferManager.instance().getActiveDownloads();
+        int uploads = TransferManager.instance().getActiveUploads();
+
+        EngineService.updatePermanentStatusNotification(
```
aldenml: Can we move the method `updatePermanentStatusNotification` to this class?

aldenml:
Hi @grzesiekrzaca, check my latest comments. After that, I'm willing to merge the PR and continue with https://github.com/frostwire/frostwire/pull/140 (after rebase).

grzesiekrzaca:
Updated with refactoring - seems nothing exploded. (I force-pushed my changes after rebasing to master, is this the right way of doing it with rebase? (the branch after rebase was 23 ahead and 2 behind))

aldenml:
Yes, rebase is the right way, you also need to squash (to have just one commit).  And since you need one more accommodation, fix the formatting here `if(notificationUpdateDemon!=null){` and other places if necessary (IntelliJ is very good at this).

PR #147, opened 2016-05-25
[common] avoid connection leaks with OKHttpClient.

gubatron:
I kept getting the following error after making a few searches while fixing a bug on EZTV.
I've had this issue in the past, and what happens when you start seeing this, is that FrostWire simply doesn't connect to any search engine, as it seems the HTTP client is exhausted.

screen shot 2016-05-24 at 10 44 47 pm

I believe you mentioned something related to this earlier this week but you know my memory.

In any case I searched for the error and the brain took me [to this conversation on OKHttp](https://github.com/square/okhttp/issues/2158) which lead to [this documentation update](https://github.com/square/okhttp/pull/2159/files#diff-8293e59e7aba0f85474c4ad7a639a4c7R35) on `ResponseBody`.

On this patch I tried to make sure that every `ResponseBody` object created was closed quietly.

I've not been able to replicate the error so far, but I'm not sure if this is the right course of action so please take a look.

### DIFF
--- a/common/src/main/java/com/frostwire/util/http/OKHTTPClient.java
+++ b/common/src/main/java/com/frostwire/util/http/OKHTTPClient.java
@@ -71,10 +71,16 @@ public class OKHTTPClient extends AbstractHttpClient {
         byte[] result = null;
         final OkHttpClient okHttpClient = newOkHttpClient();
         final Request.Builder builder = prepareRequestBuilder(okHttpClient, url, timeout, userAgent, referrer, cookies);
+        ResponseBody responseBody = null;
         try {
-            result = getSyncResponse(okHttpClient, builder).body().bytes();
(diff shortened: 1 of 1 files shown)


gubatron:
ping @aldenml

aldenml:
This should not be necessary in okhttp3, but we are still at 2.0

gubatron:
I did see an issue, see the screenshot above, at that point FrostWire wouldn't perform searches.

And I think you're right with okhttp3, I just saw the code on the repo, and `.bytes()` and `.string()` perform a `.close()`, I haven't checked the code we have, but better be sure.

gubatron:
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/ResponseBody.java#L130

aldenml:
👍

aldenml:
already merge btw

PR #151, opened 2016-05-30
 subscription asks subscriber if can be skipped

grzesiekrzaca:


gubatron:
check it out @aldenml  :)

aldenml:
Yes, I saw it, and tried to explain my point of view in the issue

PR #158, opened 2016-06-22
Desktop buttons

marcelinkaaa:
Larger, more readable transfer filter buttons and larger filter box (with the box the only thing I was not able to do is to center the text inside? Any ideas how to do it?)
The change came with no loss in vertical space between the search section and transfer section. 

separate_screen
together_screen

### DIFF
--- a/desktop/src/com/frostwire/gui/tabs/TransfersTab.java
+++ b/desktop/src/com/frostwire/gui/tabs/TransfersTab.java
@@ -77,9 +77,9 @@ public final class TransfersTab extends AbstractTab {
     }
 
     private void initComponents() {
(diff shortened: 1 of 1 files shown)


gubatron:
ha, and I made them smaller thinking they looked bad with big font.
Let me check how this looks when running.

gubatron:
so you issue is the text on the filter box?

This is how it looks for me, I suppose that's your issue, let me see if I can help
screen shot 2016-06-22 at 5 30 47 pm

marcelinkaaa:
Yes the filter box text - should be centered but no matter what I did it was always aligned to the top of the box.

gubatron:
Fixed. Good old [Bing](https://www.bing.com/search?q=java+swing+change+theme+for+single+component&PC=U316&FORM=CHROMN) to the rescue ;)

PR #169, opened 2016-07-18
Updated roboto fonts

marcelinkaaa:
Adjusted textStyles; Moved styles with Roboto Medium font to v21 style.xml


android/res/values/styles.xml
```diff
@@ -193,10 +193,10 @@
     </style>
 
     <!-- Menu Media Type end -->
-
+    <!--v21-->
```
gubatron: Q: is this some sort of system annotation that android will use? or just documentation?

marcelinkaaa: This is just a documentation for me so we know that this particular style is used separately for lollipop and later in styles.xml(v21). So that when you update the style you update it in the other version as well.

aldenml:
I'm worried about to much duplication of markup. If the difference is just the font, you should be able to factor out that particular property and keep all the rest in the regular xml.

marcelinkaaa:
The difference is just the font but when I tried using just the font in v21 the rest of the styles did not fall back on the regular style xml. 
In any case from all the styles we have this is the only thing we will have for a long time. I went through all. Let me know if we can keep it.

aldenml:
I would like to explore my suggestion, I believe it's possible. I understand what's happening to you, I just don't know exactly how to do it at this moment.

marcelinkaaa:
I'll try to look at it more in the morning. Out of the house now. :)

aldenml:
Not a problem as it is, and not worth the refactor at this moment, merging this.

PR #174, opened 2016-07-28
[android] adjusted paddings and colors

marcelinkaaa:
Small final adjustments: 
Adjusted padding so that the ad is fully visible on smaller screens, the last part was cut on some hdpi devices.
Adjusted colors of the ad-free border and text.


android/res/layout/activity_buy.xml
```diff
@@ -45,18 +45,19 @@
                      android:layout_gravity="center|right"
                      android:contentDescription="@null"/>
 
+
```
aldenml: don't mind the extra space :) (not important)

Issue #170, opened 2016-07-20
[desktop] Make Send to Friend Window appear on the same desktop as the main window ($15 bounty)


gubatron:
If you're on a multiple monitor setting it's really annoying to have the "send to friend" window appear on the main window if you've moved FrostWire to the secondary screen.

To replicate you will need 2 monitors:
- Open FrostWire
- Move main window to second monitor
- Drag and drop a file on FrostWire, say yes to the question, the "Send to Friend" window will appear on the first monitor, not the second one.

gubatron:
ping @alejandroarturom @votaguz @X-NicON

gubatron:
almost there, now just center the window before it appears, using as frame of reference the coordinates of the frame you gave it. simple arithmetic

PR #181, opened 2016-08-24
[desktop] Adding Main app frame as parameter to ShareTorrentDialog constructor

votaguz:
This pull request is intended to solve #170 .


gubatron:
good progress, however, the ShareDialog window does not appear centered with respect to the main window, like the dialog that asks before. Almost there.

gubatron:
sexykiss

gubatron:
Fixes issue #170

PR #184, opened 2016-09-05
[ANDROID] New Search Engine: LimeTorrents.

alejandroarturom:


common/src/main/java/com/frostwire/search/SearchManager.java
```diff
@@ -424,7 +425,7 @@ public SearchPerformer newPerformer(long token, String keywords) {
     public static final SearchEngine LIMETORRENTS = new SearchEngine("LimeTorrents", AppSettings.SEARCH_LIMETORRENTS_ENABLED) {
         @Override
         public SearchPerformer newPerformer(long token, String keywords) {
-            return new TorrentDownloadsSearchPerformer("www.limetorrents.cc", token, keywords, DEFAULT_SEARCH_PERFORMER_TIMEOUT);
+            return new LimeTorrentsSearchPerformer("www.limetorrents.cc", token, keywords, DEFAULT_SEARCH_PERFORMER_TIMEOUT);
```
aldenml: at some point we should have a type-based solution to avoid this mistake

PR #185, opened 2016-09-05
[Common] Minor. Typo.

alejandroarturom:
Minor. Search Engine Name doesn't need to have the domain name

### DIFF
--- a/common/src/main/java/com/frostwire/search/SearchManager.java
+++ b/common/src/main/java/com/frostwire/search/SearchManager.java
@@ -415,7 +415,7 @@ public final class SearchManager {
         }
     };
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/frostwire/search/SearchManager.java
```diff
@@ -415,7 +415,7 @@ public SearchPerformer newPerformer(long token, String keywords) {
         }
     };
 
-    public static final SearchEngine TORRENTDOWNLOADS = new SearchEngine("TorrentDownloads.me", AppSettings.SEARCH_TORRENTDOWNLOADS_ENABLED) {
```
gubatron: hmm, we gotta be careful with this..., this might be the way it was named on update.frostwire.com/android, let's see

gubatron: ok, it's good:
```
 "TorrentDownloads":true,
```

PR #193, opened 2016-09-24
Settings style update

marcelinkaaa:
Hi guys, 
It started with styling of the newly added toggle all settings... then I decided to add two new layouts - one for the toggle checkbox header and one for - well everything else. This way future styling of any elements in the settings screen will be much more easier and straight forward. Take a look and let me know if its ok.

I started off on a wrong foot but by doing it I learned quite a bit about preferences. Would love to take a try at getting rid of some deprecated items, using a PreferenceFragment instead, and setting the preferences for the use on tablets and larger screens. For that last one I will try to send mockups of what I was thinking tomorrow.


aldenml:
You know, it's not that easy to debug styles without actually testing :D 
Please change these lines `...g="utf-8"?><!--` to have the `<!--` in a new line.

aldenml:
I would love to have a solution with `PreferenceFragment`, but I tell you that the amount of logic in the current `SettingsActivity` is just insane. I will give you more feedback when I test this in the phone

gubatron:
why the name change on CheckBoxPreferenxe to CheckboxSeedingPreference? were those checkboxes only used for the seeding setting?

android/src/com/frostwire/android/gui/activities/ToggleAllSearchEnginesPreference.java
```diff
@@ -60,15 +60,15 @@ public CheckBox getCheckbox() {
...
         titleView.setText(getTitle());
-        view.findViewById(R.id.checkbox_preference_summary).setVisibility(View.INVISIBLE);
 
-        checkbox = (CheckBox) view.findViewById(R.id.checkbox_preference_checkbox);
+        checkbox = (CheckBox) view.findViewById(R.id.view_preference_checkbox_header_checkbox);
         checkbox.setVisibility(View.VISIBLE);
         checkbox.setClickable(true);
         checkbox.setChecked(areAllEnginesChecked());
 
+
```
gubatron: does it look better now horizontally with these changes? love it when you anticipate my next fix

marcelinkaaa: it does. And as to the name change for the CheckBoxPreference to seeding - yes it was used only for seeding. For everything else you could just use a LinearLayout with "@+android:id/widget_frame" and if you used CheckBoxPreference or SwitchPreference etc it would just place the correct thing into that layout. 
The CheckboxSeedingPreference uses a checkbox specifically so it could not be used on all other preferences and you use the id of that checkbox specifically for the logic related to seeding.

gubatron: awesome, I'll test this tomorrow morning

android/build.gradle
```diff
@@ -24,7 +24,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.3'
+        classpath 'com.android.tools.build:gradle:2.2.0'
```
aldenml: this will not work with intellij

gubatron: too late, android studio is like twice as fast

gubatron: you made me do it!!! :)

aldenml: OK :D I will test it then

gubatron: oh, and I learned how to export the keymaps!

Go to `File -> Export Settings`, select only the keymaps, it'll create a .jar, then in Android Studio, `File -> Import Settings`.

gubatron: oh, and I learned how to export the keymaps!

Go to `File -> Export Settings`, select only the keymaps, it'll create a .jar, then in Android Studio, `File -> Import Settings`.

PR #201, opened 2016-10-11
[Android] French Translation. Added Missing Strings

alejandroarturom:
I added missing strings and translated it. Obviously it's not a perfect translation, and if is there any french native speaker reading this, please give us your feedback !


gubatron:
wow, that was a lot, Thank you so much.

I believe these are the ones we'll do next:
```
- German
- Russian
- Mandarin
- Hindi
- Dutch (NL - Netherlands)
```

PR #203, opened 2016-10-13
[Android] German Translation. Added missing lines.

alejandroarturom:
German translation. Added all missing lines. It's not a perfect translation, so if any German native speaker is reading this, please help us :)


gubatron:
Thank you so much!

Issue #233, opened 2016-11-03
[android] Problems with the state of playback in My Files lists - not refreshed when populating views


grzesiekrzaca:
Android 6.0.1 Samsung S5

1) Navigate to My Files > My Music
2) Begin playback of a file
3) Press back button to minimize the player
4) Scroll down
Is:
Other files (cyclically one roughly one screen length down) have 'stop' button 
Expected:
Other files have 'play' button

Issue #226, opened 2016-11-02
[android] My Files > My Ringtones empty context menu when multiple files are selected


grzesiekrzaca:
Android 6.0.1 Samsung S5

1) Navigate to my ringtones
2) Check 2 ringtone files 
3) Hold to open context menu 
Is:
Empty menu
Expected:
Populated context menu
 
Alternative repro:
1) Navigate to my ringtones
2) Check 1 ringtone file
3) Navigate to my music
4) Check  at least one file
5) Navigate to my ringtones 
6) Hold to open context menu 
Is:
Empty menu
Expected:
Populated context menu

gubatron:
Good find, this is happening because of changes in android that don't allow us anymore to delete ringtones across all devices.

Plan of action:

1. Hide the checkboxes capability when the user has selected the ringtone media type.
2. When a ringtone is touched (on the right hand side), or long pressed, show an action menu that contains the following actions: "Open", "Use as ringtone"

PR #234, opened 2016-11-04
[android] Issues with ringtone playback (#225, #226, #233)

### LINKED ISSUE #225: [android] Fix My Files > Ringtone playback
When an user tries to preview the sound of a Ringtone file in "My Files > Ringtones", FrostWire should not try and open the music player screen, it should just play the ringtone once on the spot.

Details to consider:
 - A song might playing from the music player. That song should be stopped (or perhaps paused, as the user might want to go back to the song that was being played)
 - We should not let the user play 2 ringtones at once
 - Ideally we should change the play button on the ringtone in question and ignore all input on that view while the ringtone is being played.

Tips:
 I took a look at the code last week and I believe the `MediaType` related to the ringtone can be of help on how to handle the playback logic.

grzesiekrzaca:
this fixes 
#225
#226 
#233 (actually that is just a one line fix in the BrowseThumbnailImageButton and was needed in the above issues anyway)

### DIFF
--- a/android/apollo/src/com/andrew/apollo/IApolloService.aidl
+++ b/android/apollo/src/com/andrew/apollo/IApolloService.aidl
@@ -9,6 +9,8 @@ interface IApolloService
     void stop();
     void pause();
     void play();
+    void playSimple(String path);
+    void stopSimplePlayer();
     void prev();
     void next();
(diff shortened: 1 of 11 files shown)


gubatron:
I've rebased on another branch, merging from there.

PR #254, opened 2016-11-16
[android] moving creation of shortcut on home screen to asynchronous task

grzesiekrzaca:
As per @aldenml request making this part separate

### DIFF
--- a/android/apollo/src/com/andrew/apollo/ui/activities/ProfileActivity.java
+++ b/android/apollo/src/com/andrew/apollo/ui/activities/ProfileActivity.java
@@ -41,6 +41,7 @@ import com.andrew.apollo.utils.*;
 import com.andrew.apollo.widgets.ProfileTabCarousel;
 import com.andrew.apollo.widgets.ProfileTabCarousel.Listener;
 import com.frostwire.android.R;
+import com.frostwire.util.Ref;
 
 import java.lang.ref.WeakReference;
 
@@ -336,7 +337,7 @@ public class ProfileActivity extends BaseActivity implements OnPageChangeListene
                 // screen. Definitely one of my favorite features.
                 final String name = isArtist() ? mArtistName : mProfileName;
                 final Long id = mArguments.getLong(Config.ID);
-                ApolloUtils.createShortcutIntentAsync(name, mArtistName, id, mType, new WeakReference<Activity>(this));
+                ApolloUtils.createShortcutIntentAsync(name, mArtistName, id, mType, Ref.weak((Activity) this));
                 return true;
             }
             case R.id.menu_shuffle: {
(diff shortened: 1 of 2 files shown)


android/apollo/src/com/andrew/apollo/utils/ApolloUtils.java
```diff
@@ -144,47 +145,66 @@ public static final ImageFetcher getImageFetcher(final Activity activity) {
...
+                                    context,
+                                    context.getString(R.string.could_not_be_pinned_to_home_screen, displayName),
+                                    AppMsg.STYLE_ALERT).show();
+                        }
+                    }
+                };
+                context.runOnUiThread(postExecute);
+            }
+        };
+        Engine.instance().getThreadPool().submit(task);
```
gubatron: :)

android/apollo/src/com/andrew/apollo/utils/ApolloUtils.java
```diff
@@ -144,47 +145,66 @@ public static final ImageFetcher getImageFetcher(final Activity activity) {
...
+                                    AppMsg.STYLE_CONFIRM).show();
+                        } else {
+                            AppMsg.makeText(
+                                    context,
+                                    context.getString(R.string.could_not_be_pinned_to_home_screen, displayName),
+                                    AppMsg.STYLE_ALERT).show();
+                        }
+                    }
+                };
+                context.runOnUiThread(postExecute);
```
gubatron: perhaps here we should check if this context is still alive.

Also, to avoid leaking it, perhaps we should work with `WeakReference<Context>` instead.

Just thinking of what happens when things take long and the phone is rotated, goes to sleep, etc.

gubatron: so this part would look more like

```
if (Ref.alive(contextRef)) {
    contextRef.get().runOnUiThread(postExecute);
}
```

gubatron: what do you think @aldenml ?

aldenml: I rather pass an `ApplicationContext` as the parameter to avoid all that

grzesiekrzaca: Well, this will hold it alive, right?
I tested that with Thread.sleep(10000) to simulate a long running image get. 
Backed out of the activity/ home buttoned and locked and no problems.

aldenml: I think the context is pinned by the call, but not necessary "alive" in the sense that the activity could be destroyed, but no necessarily garbage collected. And this happens based on memory pressure, not time based.

grzesiekrzaca:
I switched to using WeakReference.
Cannot use application context here without changing both messages (it needs activity, or it crashes) and old image cache (which seems counterproductive)

android/apollo/src/com/andrew/apollo/ui/activities/ProfileActivity.java
```diff
@@ -334,7 +336,7 @@ public boolean onOptionsItemSelected(final MenuItem item) {
                 // screen. Definitely one of my favorite features.
                 final String name = isArtist() ? mArtistName : mProfileName;
                 final Long id = mArguments.getLong(Config.ID);
-                ApolloUtils.createShortcutIntent(name, mArtistName, id, mType, this);
+                ApolloUtils.createShortcutIntentAsync(name, mArtistName, id, mType, new WeakReference<Activity>(this));
```
aldenml: `Ref.weak`?

Issue #252, opened 2016-11-16
[android] Lockscreen playback notification has no controls


grzesiekrzaca:
Samsung S5 Android 6.0.1

![device-2016-11-16-122508](https://cloud.githubusercontent.com/assets/2766994/20345556/fcc9826c-abf7-11e6-9270-118bfda346ca.png)

Ukryta zawartość = hidden content

gubatron:
what do you suggest we do here?

Should we just make it unavailable when the screen is locked, thus protecting the user's privacy?

gubatron:
CC: @marcelinkaaa

grzesiekrzaca:
Well, I'd at least add the pause button - The music is still playing (So protecting the user's privacy by not displaying what is playing is weird). We need to unlock the screen to change that (and that's a hassle if someone has an elaborate swipe pattern).

marcelinkaaa:
@grzesiekrzaca Check your settings. Some content like messages is hidden on lock screen for privacy. It could be similar with players. On my phones I am getting the standard player we have (although I agree the player itself could have more options - I will upload what I mean to your ticket about that).
![screenshot_20161116-073816](https://cloud.githubusercontent.com/assets/2339972/20347631/1a93df40-abd0-11e6-9aee-05eb399fae82.png)

grzesiekrzaca:
Yeah, I guess it's the android version thing and they changed something in permissions again. I can see other players on my lock screen just fine. Which android do you have?

marcelinkaaa:
HTC M8 and LG Nexus - both running Android 6

grzesiekrzaca:
Hmm, weird. All other players (default Samsung one and MX Player)  have normal controls in the notification on the lockscreen but FW notifications have their contents hidden.

marcelinkaaa:
I just downloaded bunch of other players and all of them are open on lockscreen including FW. 
Will ask others what they have. Maybe its phone related only or maybe something with permissions on our end?

gubatron:
Thank you for noticing, suggesting and fixing this issue @grzesiekrzaca 
![thanks-bow](https://cloud.githubusercontent.com/assets/163977/20410964/c0b9be14-acdc-11e6-99e9-49c44dcdc830.gif)

Issue #251, opened 2016-11-16
[android] notification music playback controls are poor when notification cannot expand 


grzesiekrzaca:
Samsung S5 Android 6.0.1

Repro:
1) Have many permanent notifications 
2) Start music playback
3) Press the home screen button
4) See the notifications
Is:
![untitled](https://cloud.githubusercontent.com/assets/2766994/20345414/53843b0c-abf7-11e6-8891-31c28f5adaf1.png)
Expected:
Something better than forward track and close?

gubatron:
I'll let Marcelina decide what else makes sense to add here.

I suppose there's plenty of room there to at least add a back and a pause button.

There's also the issue that it has a white background when in its simple form, and when expanded it has a black background, perhaps you also opened that ticket.

marcelinkaaa:
I will look into the background color changes. 

So few options: 
Small player:
- we should at least have 1, 2
- in most players - clicking on 'x' pauses the song - and gets rid of the notification. It would be great if that 'x' was used the same way as long press on 'play' in the player - meaning it would stop the player entirely and exit the notification. 
- if we close the player by swiping the notification to the right of the screen, we could also do what is on screenshot nr. 3. Or replace the track back for 'heart' to favorite the song (if we have the data that would suggest that users like favoriting the songs on the player; its the action that is used often)

Big player:
- change the color of the background nr.4
- 5 is just something I seen once - I dont think its necessary so would stick with 4 unless its super easy to accomplish
- 6 - would prefer that one if we can prove that favorites is used widely by the users. 

1
![1](https://cloud.githubusercontent.com/assets/2339972/20351338/6e41f824-abe0-11e6-9dad-d87143f010b7.png)
2
![2](https://cloud.githubusercontent.com/assets/2339972/20351339/6e44c752-abe0-11e6-822a-9a37dc8dbf21.png)
3
![3](https://cloud.githubusercontent.com/assets/2339972/20351341/6e45d41c-abe0-11e6-85d2-b015e51b9595.png)
4
![4](https://cloud.githubusercontent.com/assets/2339972/20351343/6e478794-abe0-11e6-83c9-c1cb29f44695.png)
5
![5](https://cloud.githubusercontent.com/assets/2339972/20351340/6e45e952-abe0-11e6-950a-658e33c8869a.png)
6
![6](https://cloud.githubusercontent.com/assets/2339972/20351342/6e465a2c-abe0-11e6-8165-c7ba6ee13bdb.png)

[...]

PR #257, opened 2016-11-17
[android] Make control notification visible

grzesiekrzaca:
#252 
This will allow the notification contents to be visible on lock screen when notification privacy is up. On my phone (S5 6.0.1) at least this has a side effect of the fact that the notification in notification area is now expanding (previously lots of ongoing notifications blocked that expansion)
so it affects #251

### DIFF
--- a/android/apollo/src/com/andrew/apollo/NotificationHelper.java
+++ b/android/apollo/src/com/andrew/apollo/NotificationHelper.java
@@ -97,6 +97,7 @@ public class NotificationHelper {
                 .setSmallIcon(getNotificationIcon())
                 .setContentIntent(pendingintent)
                 .setPriority(Notification.PRIORITY_DEFAULT)
+                .setVisibility(Notification.VISIBILITY_PUBLIC)
                 .setContent(mNotificationTemplate)
                 .build();
         // Control playback from the notification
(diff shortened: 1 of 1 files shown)


aldenml:
This is API 21.

As a general note (for all of us), this is precisely the reason I didn't want to up the SDK api. I suppose you get some warnings in the build process, or not?

grzesiekrzaca:
AFAIK Pre 21 devices will not show the notification on the lock screen, and this will not have an effect on the normal notification (at least that how NotificationCompat.Builder should work)
I don't have an old device at the moment i'll check on an emulator.
Or do you just want an if(Build.VERSION.SDK_INT >= ) to be extra safe?

grzesiekrzaca:
Tested on emulator API 19, no issues.

aldenml:
I rather no access/touch any API beyond 16. Is this the only solution to the issue?

grzesiekrzaca:
No, that's "new" functionality. 
They messed around with allowing custom stuff on the lock screen between 4.1 (or close enough) and 5.0 (API 21) when they removed it and moved the interaction with apps to notifications. So they introduced various levels of visibility (public (show all),private (show notification but protect content),secret (don't show the notification)) to accommodate that.

aldenml:
Shall we use `NotificationCompat.Builder` instead?

grzesiekrzaca:
I don't understand. We use NotificationCompat.Builder

aldenml:
right, then you should put a private constant with the value `Notification.VISIBILITY_PUBLIC` and use that instead

PR #263, opened 2016-11-18
[android] asynchronous blurring

grzesiekrzaca:


android/src/com/frostwire/android/util/ImageLoader.java
```diff
@@ -148,6 +150,34 @@ private ImageLoader(Context context) {
...
+        @Override
+        public Bitmap transform(Bitmap source) {
+            Bitmap blurred = BitmapUtils.createBlurredBitmap(source);
+            source.recycle();
+            return blurred;
+        }
+
+        @Override
+        public String key() {
+            return "blurred";
```
aldenml: did you find how the internal cache key is generated?

aldenml:
The two new methods seems very particular to be included in `ImageLoader`, maybe you can improve the API to pass a transformation to be used in the particular place it's used.

aldenml:
So, the idea here is that you will cache the blur version of the images. Is it that important? aren't you worried that the blur versions will evict the more important ones, or create a constant eviction/rotation overall?

grzesiekrzaca:
Well, my reasoning is that blurring is a costly operation and that should be cached. We don't blur that often to pose problems for the cache but if the user navigates back to the same screen he won't have to wait for blurring. It will be there (assuming he wasn't all over the app and the cached bitmap got evicted)

aldenml:
what if we don't blur at all? I know is a design decision, @marcelinkaaa what do you think if we get a blurred image that kind of fill reasonable well any background/contrast?

marcelinkaaa:
Are you guys talking about the blur just like on the first attached image? 
Then - we should keep it. In the album section we should show entire album cover and blur makes the width of the screen fill in nicely. 
But there is a little inconsistency on how we show things later - look at image 2 & 3.  On one the cover from the either left or right to the picture on the current screen is placed under a black overlay, on the other picture it is not. 

screenshot_20161119-100249
screenshot_20161119-100257
screenshot_20161119-100311

aldenml:
@grzesiekrzaca I think we should tackle the inconsistency first, maybe that could give me more arguments to remove the blur.

android/src/com/frostwire/android/util/ImageLoader.java
```diff
@@ -148,6 +149,20 @@ private ImageLoader(Context context) {
...
+
+    public void loadAndTransformWithAlternative(final Uri primaryUri, final Uri secondaryUri, final Transformation transformation, final ImageView mPhoto) {
+        picasso.load(primaryUri).fit().transform(transformation).into(mPhoto, new Callback.EmptyCallback() {
+            @Override
+            public void onError() {
+                loadAndTransform(secondaryUri, transformation, mPhoto);
+            }
+        });
+    }
+
```
gubatron: I'd refactor to avoid repeating the core logic with picasso:

```java
public void loadAndTransform(final Uri primaryUri, final Uri secondaryUri, final Transformation transformation, final ImageView imageView) {
    Callback.EmptyCallback callback = new Callback.EmptyCallback() {
        @Override
        public void onError() {
            if (secondaryUri != null) {
                loadAndTransform(secondaryUri, transformation, imageView);
            }
        }
    };
    picasso.load(uri).fit().transform(transformation).into(imageView, callBack);
}


public void loadAndTransform(Uri uri, Transformation transformation, ImageView imageView) {
    loadAndTransform(uri, null, transformation, imageView);
}

```

aldenml:
I'm still very cautious on this change, big images in the same cache used for our list adapters will impact the performance heavily. Can you enable the picasso debug feature and tell me if you see a meaningful difference in the amount of eviction for the thumbnails images?

gubatron:
If the eviction number is high, think we should have multiple cache instances (saving to different locations, based on their names) something like:

`ImageLoader.getInstance("myfiles", context)`
`ImageLoader.getInstance("search", context)`
`ImageLoader.getInstance("apollo", context)`

and ideally be able to configure the maximum disk storage of each so we don't take too much space.

grzesiekrzaca:
I think multiple instances would mean smaller caches and even more evictions. (Talking about the "worst cache" even if others are more stable) I'll look at the cache performance tomorrow and propose a course of action.

grzesiekrzaca:
About the cache (values were taken from an S5):
Memory cache size is 19173961 bytes (~18,2 MB)
Disk is 52428800 bytes (50MB - max disk cache size)
When it comes to small images (thumbnails in My Files) the cache can hold hundreds of them
example: `image://album/13-resize:96x96- 96 x 96 at 36864` so just  36KB
but when it comes to My Music things are not that great. There is no resize set so whatever the size it gets into the cache (That makes the disk portion of the cache redundant. And can lead to problems with super large images (OOMs) [on the main problem branch I added a hard 768x768 clamp. Now I'm thinking about computing that value depending on the size of the device(same could be done for thumbnails)]). This is usually around 200x200 and 600x600 (at least for me):
`image://album/27- 200 x 200 at 160000` ~156 KB
`image://album/79- 600 x 600 at 1440000` ~1,37 MB (this is displayed in a 540 x 540 square so an overkill even without consideration that it's just 3cm of screen and probably even less detailed image would look good )
but there are outliers like 
`image://album/1- 1072 x 1072 at 4596736` ~4,4 MB
and
`image://album/13- 1460 x 1460 at 8526400` ~8,13 MB
the blurred image:
`image://artist/Marcin Przybyłowicz-resize:1080x589-blurred- 1080 x 589 at 2544480` ~2,42 MB 
So at the moment just going to the My Music (artists/albums) and scrolling through a page or two fills up the cache.
Also, the search screen images are quite big and quickly get thrown out of memory cache.
`http://static.frostwire.com/images/overlays/jyant_love_overdose_overlay_1.jpg-resize:1080x572- 1080 x 572 at 2471040` ~2,36 MB

Speaking of the blurred bitmap there are 2 solutions (if we don't want it in the cache):
1) Don't store it in the cache. :)
2) Store smaller bitmap in the cache. (It's blurred so upscaling would make the blur stronger)
`image://artist/Marcin Przybyłowicz-resize:540x295-blurred- 540 x 295 at 637200` ~622,2 KB
screenshot_4
screenshot_3

@aldenml

gubatron:
option 3: get rid of effing blurring :p

@aldenml should be answering soon

grzesiekrzaca:
hehe 
well yeah, that's an option too :) but I was trying to stay true to what @marcelinkaaa has said: 
> "Then - we should keep it.".

aldenml:
@grzesiekrzaca thanks for the analysis and numbers, question: Do you think is it possible (and good idea) to have a way to specify if an image will be cached or not, but still use `ImageLoader`?

@marcelinkaaa there are multiple technical benefits in no having the blur effect. Do you think an alternative solution (like a fixed background image) is acceptable from the UI point of view?

gubatron:
If calculating this blur on a downscaled version of the image doesn't cost much in terms of CPU/load time, let's just calculate it every time.

If it's expensive, then let's store a very small version of it, like you say, the smaller the more accentuated it will be and it won't cause too many evictions.

I'd also try to have multiple cache buckets for different purposes, kind of what we used to have, but using a single caching component (ImageLoader), this way when you browse my music, you don't get rid of things like the promo images on the home page.

also for search results, let's try to make sure we downsize all images coming in from some search results (like the youtube ones)

aldenml:
agree with on the fly calculation, and if possible use RenderScript for that, it should be off CPU and faster (API >= 17)

gubatron:
here's an old example w/renderscript, might need some fixes
https://futurestud.io/tutorials/how-to-blur-images-efficiently-with-androids-renderscript

grzesiekrzaca:
Disabled caching and switched to renderscript

android/build.gradle
```diff
@@ -79,7 +79,8 @@ android {
 
     defaultConfig {
         versionName project.ext.versionName
-
+        renderscriptTargetApi 16
```
aldenml: I think this needs `android:minSdkVersion="17"` right? This will need a final decision from @gubatron

aldenml:
Now I'm starting to feel more comfortable with the change

android/build.gradle
grzesiekrzaca: raw renderscript does need it but setting this to lower version enables support.v8 version of renderscript (alternatively check version and when >= 17 use renderscript and old blur function otherwise)

aldenml: good, didn't know

android/src/com/frostwire/android/util/ImageLoader.java
```diff
@@ -148,6 +149,30 @@ private ImageLoader(Context context) {
         picasso.setIndicatorsEnabled(false);
     }
 
+    public void loadAndTransform(final Uri primaryUri, final Uri secondaryUri, final Transformation transformation, final ImageView imageView, final boolean cache) {
```
grzesiekrzaca: What do you mean by ""Filter" interface"?

aldenml: create an interface inside ImageLoader that mimics the `Transformation` from picasso. Feel free to put another name (like Filter) and then use that interface for the ImageLoader API.

android/apollo/src/com/andrew/apollo/widgets/CarouselTab.java
```diff
@@ -14,17 +14,19 @@
...
 import android.widget.ImageView;
 import android.widget.TextView;
+
 import com.andrew.apollo.cache.ImageFetcher;
 import com.andrew.apollo.utils.ApolloUtils;
 import com.andrew.apollo.utils.BitmapUtils;
 import com.andrew.apollo.utils.MusicUtils;
 import com.frostwire.android.R;
+import com.frostwire.android.util.ImageLoader;
+import com.squareup.picasso.Transformation;
```
aldenml: cleanup here

android/src/com/frostwire/android/util/ImageLoader.java
```diff
@@ -148,6 +149,46 @@ private ImageLoader(Context context) {
         picasso.setIndicatorsEnabled(false);
     }
 
+    public void loadAndFilter(final Uri primaryUri, final Uri secondaryUri, final Filter filter, final ImageView imageView, final boolean cache) {
```
aldenml: what do you think of naming this just `load` like the rest?

android/src/com/frostwire/android/util/ImageLoader.java
```diff
@@ -148,6 +149,46 @@ private ImageLoader(Context context) {
...
+        }
+    }
+
+
+    public void loadAndFilter(Uri uri, Filter transformation, ImageView imageView, boolean cache) {
+        loadAndFilter(uri, null, transformation, imageView, cache);
+    }
+
+    private Transformation packInTransformation(final Filter filter) {
+        return new Transformation() {
```
aldenml: can't be this inlined in the method, similar to the callback?

aldenml:
after addressing my latest comments, I'm ready to merge this

PR #275, opened 2016-12-02
[android] remove useless file and use only one layout file for files list

grzesiekrzaca:
This file is almost an exact copy of:
`layout\view_browse_thumbnail_peer_list_item.xml` 
the only difference is a frame with a button which only use is to be hidden in the code
```
ImageButton preview = findView(view, R.id.view_browse_peer_list_item_button_preview);
preview.setVisibility(View.GONE);
```
and the id of the custom imagebutton


aldenml:
well, it used to be a little different in the layout structure, I welcome the simplification

android/src/com/frostwire/android/gui/adapters/menu/FileListAdapter.java
```diff
@@ -90,7 +89,7 @@ public byte getFileType() {
 
     @Override
     protected final void populateView(View view, FileDescriptorItem item) {
-        if (getViewItemId() == R.layout.view_browse_thumbnail_peer_list_item) {
+        if (fileType == Constants.FILE_TYPE_PICTURES || fileType == Constants.FILE_TYPE_VIDEOS || fileType == Constants.FILE_TYPE_APPLICATIONS || fileType == Constants.FILE_TYPE_AUDIO) {
```
aldenml: do you mind putting this in a method? (something like `hasThumbnailView`)

android/src/com/frostwire/android/gui/adapters/menu/FileListAdapter.java
```diff
@@ -90,7 +89,6 @@ public byte getFileType() {
 
     @Override
     protected final void populateView(View view, FileDescriptorItem item) {
-        byte fileType = item.fd.fileType;
```
gubatron: oh, you had added it there

gubatron:
after the `hasThumbnailView(byte fileType)` has been added it's ACK for me.

android/src/com/frostwire/android/gui/adapters/menu/FileListAdapter.java
gubatron: I also wonder if that method should be the other way around logically (for less code)
```
# pseudo code
return fileType not in (torrents, docs)
```

grzesiekrzaca:
Sorry Sourcetree hang up and I haven't noticed it until now (I was away)

gubatron:
Thank you for this @grzesiekrzaca 
thanks_dalai_lama

PR #284, opened 2016-12-09
[android] adding material theme to BuyActivity.

marcelinkaaa:
Don't want to change any styles yet. Just wanted to extend AbstractActivity2 and set up a new toolbar. Let me know if its ok. 
Removed toolbar when in InterstitialMode.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -118,7 +118,7 @@
         <activity android:name=".gui.activities.SettingsActivity"
                   android:label="@string/activity_preferences_name" android:theme="@style/Preferences"/>
         <activity android:name="com.frostwire.android.gui.activities.BuyActivity"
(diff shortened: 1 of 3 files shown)


aldenml:
In principle I don't see nothing wrong, but maybe the override of `onCreate` can be removed.

aldenml:
does this toolbar pick the random pitch in the title?

marcelinkaaa:
Hmmm I thought it only did so when in interstitial mode. Just came back will check in a bit.

aldenml:
much better, any visual change in L&F? screenshot?

marcelinkaaa:
No visual changes in layout or function... Just adding toolbar. The title together with a random pitch is shown in settings -> remove ads as well as in interstitial version. 

screenshot_20161209-141355
screenshot_20161209-141918

PR #286, opened 2016-12-09
[android] fix list refresh

### LINKED ISSUE #285: [android] My files when switching filtered lists unfiltered list is shown for a split second
1) Filter file list so that top element is not visible or no elements are visible (this step is for easier observation)
2) Change to another file type and back again 
Is:
After switching, the list with unfiltered file list is shown and then after the filtering is applied the filtered list is shown (it takes a split second)
Expected:
Only the filtered list is shown

grzesiekrzaca:
fixes #285


aldenml:
I'm not sure about this. Not clear to me if this only improve a weak/faulty feature execution and this is in part the reason `BrowsePeerFragment` is in such a bad state. I rather go to the root, just an opinion.

gubatron:
I think this fix is overly complicated. I'm submitting a refactor that's way more readable.

PR #288, opened 2016-12-10
Preferences design update

marcelinkaaa:
@aldenml - Update to the layout of preferences as discussed with @gubatron, added toolbar to SettingsActivity and its nested screens, added material theme, and adjusted padding/dividers/text sizes of layouts used according to guidelines. 

All preferences are working as before, no functionality changes made. 
Next step: add/replace talked about preferences ('Download on WiFi only' etc.), change number picker dialog as per discussion with grzesiek... Clean up settings code - a lot of it is deprecated already.

### DIFF
--- a/android/res/layout/view_preference_buy_ads.xml
+++ b/android/res/layout/view_preference_buy_ads.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+/*
+ * Created by Angel Leon (@gubatron), Alden Torres (aldenml),
+ *            Marcelina Knitter (@marcelinkaaa)
+ * Copyright (c) 2011-2017, FrostWire(R). All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+-->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:orientation="vertical">
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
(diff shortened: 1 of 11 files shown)


android/res/layout/view_preference_buy_ads.xml
```diff
@@ -0,0 +1,84 @@
...
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_gravity="center_vertical"
+                android:layout_marginLeft="10dp"
+                android:background="@drawable/ad_menu_item_background_filled"
+                android:paddingBottom="2dp"
+                android:paddingLeft="5dp"
+                android:paddingRight="5dp"
+                android:paddingTop="2dp"
+                android:text="PRO"
```
aldenml: this text should be in the resources

android/res/layout/view_preference_simple_action_padding.xml
```diff
@@ -0,0 +1,76 @@
...
+        <LinearLayout
+            android:id="@+android:id/widget_frame"
+            android:layout_width="wrap_content"
+            android:layout_height="match_parent"
+            android:gravity="center_vertical"
+            android:orientation="vertical" />
+
+    </LinearLayout>
+
+    <ImageView
```
aldenml: this one, right?

android/src/com/frostwire/android/gui/activities/SettingsActivity.java
```diff
@@ -98,12 +99,21 @@ public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         addPreferencesFromResource(R.xml.application_preferences);
 
-        getListView().setPadding(20, 0, 20, 0);
+        LinearLayout root = (LinearLayout) findViewById(android.R.id.list).getParent().getParent().getParent();
```
aldenml: hehehe, what is this hack about?...let's find the **right** way to do this, calling `getParent()` three times is very weak at best.

android/src/com/frostwire/android/gui/activities/SettingsActivity.java
```diff
@@ -657,12 +659,32 @@ protected void onPostExecute(Void result) {
...
+
+        return false;
+    }
+
+    public void setUpNestedScreen(PreferenceScreen preferenceScreen) {
+        final Dialog dialog = preferenceScreen.getDialog();
+
+        Toolbar nestedSettingsToolbar;
+
+        LinearLayout root = (LinearLayout) dialog.findViewById(android.R.id.list).getParent();
```
aldenml: hmm...similar hack solution here.

aldenml:
Can this PR be divided in two? I'm thinking of:
1- Switching to the use of toolbar and new theme, with a less assumption of the layout structure.
2- Making all the styles and xml changes.

What do you think?

PR #290, opened 2016-12-10
[android] added toolbar and new material theme to Preferences

marcelinkaaa:
@aldenml here is just the addition of new theme and a toolbar. No layout or bigger style changes.
There is an empty space over the first preference (BitTorrent connection status) - it is due to the material theme in connection to <PreferenceCategory> and it's title - which we don't use in the old layout but the space remains. 
In the new layout this will not be an issue. But let me know if you decide to keep this layout for a longer while - I will look into a styling/layout solution to get rid of it.

### DIFF
--- a/android/src/com/frostwire/android/gui/activities/SettingsActivity.java
+++ b/android/src/com/frostwire/android/gui/activities/SettingsActivity.java
@@ -18,7 +18,6 @@
 
 package com.frostwire.android.gui.activities;
 
-import android.app.ActionBar;
 import android.app.Activity;
 import android.app.Dialog;
 import android.app.NotificationManager;
@@ -30,6 +29,8 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 import android.preference.*;
 import android.preference.Preference.OnPreferenceChangeListener;
+import android.support.v7.widget.Toolbar;
+import android.view.LayoutInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.View.OnClickListener;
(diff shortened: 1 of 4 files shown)


android/src/com/frostwire/android/gui/activities/SettingsActivity.java
```diff
@@ -121,15 +131,6 @@ public void onCreate(Bundle savedInstanceState) {
         updateConnectSwitch();
     }
 
-    private void hideActionBarIcon(ActionBar bar) {
-        if (bar != null) {
-            bar.setDisplayHomeAsUpEnabled(true);
```
gubatron: ha, so no more of this stuff with material design?

android/src/com/frostwire/android/gui/activities/SettingsActivity.java
```diff
@@ -657,12 +658,32 @@ protected void onPostExecute(Void result) {
...
+
+        return false;
+    }
+
+    public void setUpNestedScreen(PreferenceScreen preferenceScreen) {
+        final Dialog dialog = preferenceScreen.getDialog();
+
+        Toolbar nestedSettingsToolbar;
+
+        LinearLayout root = (LinearLayout) dialog.findViewById(android.R.id.list).getParent();
```
gubatron: this code right here looks a lot like the code up there... you know what that means right @marcelinkaaa ? :)

gubatron: I suggest you have a private method called:

`public void initSettingsToolbar(Context context, String title, View.OnClickListener clickListener)`

I bet you can reuse such a function for both the case above, and also for the dialog based code on the nested screen.

marcelinkaaa: So with this one if I remember correctly... from what I read Preference Screens are based on a wrapper dialog - so this was a way to capture that layout and add a toolbar to it. Do you think there is a much easier way of doing it?

aldenml: I don't know. What's the problem? That `PreferenceActivity` does not have a toolbar because the layout is not provided the same way? What about using `PreferenceFragment` and a regular activity?

aldenml:
Hi @marcelinkaaa, after some time, I finally came up with a solution with no hacks. The problem is that when you do something like the following
```java
LinearLayout root = (LinearLayout) findViewById(android.R.id.list).getParent();
Toolbar settingsToolbar = (Toolbar) LayoutInflater.from(this).inflate(R.layout.toolbar_main, root, false);
root.addView(settingsToolbar, 0); // insert at top
....
getListView().setPadding(0, 0, 0, 0);
getListView().setDivider(new ColorDrawable(this.getResources().getColor(R.color.basic_gray_dark_solid)));
getListView().setDividerHeight(1);
```
you are assuming to much of undocumented and private parts that can vary from OEMs and versions. Besides, this path don't improve the current logic, it just carry over the same...deprecated way.

Take a look at the commit https://github.com/frostwire/frostwire/commit/c83a5c01dd4f24ea243f629faf66b815f454cd5c, it just use the supported provided API and since it's based on the `PreferenceFragment` each "group" of settings will be handled separately, both in the layout/xml and the logic. I took some ideas from `PreferenceActivity`, extending the new activity to support tablets in the future should be easy. Hopefully this is a good foundation for the new settings.

Now, as you can see in master, you will have two settings in the drawer menu (easy to hide in case of release). You should be able to advance in the style part without much need of java coding, please, work with @grzesiekrzaca to handle the actual coding of each one. You can look at the old one in parallel, since both are there and only the old one is fully working.

PR #302, opened 2016-12-15
[android] make checked listener in abstract list adapter operate on items instead of views.

grzesiekrzaca:
Related to #293 
So I needed an information for the adapter that the item check state was changed so that the logic of check all box could work. Due to the fact that we can check an item without using the check box but any part of the item (on that branch in check mode) it would be hard to find the checkbox without some .getParent() conga line. I really did not need the exact item because all I needed was the info that any item was passed, so I could have passed a null but that somehow did not seem clean. So I thought why not make the adapter return the item itself rather than a CompundButton with the item as a tag.

### DIFF
--- a/android/src/com/frostwire/android/gui/dialogs/AbstractConfirmListDialog.java
+++ b/android/src/com/frostwire/android/gui/dialogs/AbstractConfirmListDialog.java
@@ -54,7 +54,7 @@ import java.util.*;
  *
  */
 abstract class AbstractConfirmListDialog<T> extends AbstractDialog implements
-        AbstractListAdapter.OnItemCheckedListener {
+        AbstractListAdapter.OnItemCheckedListener<T> {
     static final String BUNDLE_KEY_CHECKED_OFFSETS = "checkedOffsets";
     static final String BUNDLE_KEY_LAST_SELECTED_RADIO_BUTTON_INDEX = "lastSelectedRadioButtonIndex";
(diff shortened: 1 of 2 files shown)


aldenml:
I fail to see how this helps you or makes some improvement. If anything I would like to remove `OnItemCheckedListener` from `AbstractListAdapter`. Isn't the method  `AbstractListAdapter#onItemChecked` already protected?

grzesiekrzaca:
Well I needed to add a method that allows to programmatically check an item (thanks to clicking on item not on checkbox) this happens in FileListAdapter, but I need to be informed in BrowsePeerFragment of that action (for check all button). So using this listener seemed like a good idea, but I decided that it would make more sense if the listener would inform of which item was checked and not which compound button associated with this item was clicked (not really need for me as I need just the fact that there was a change in one of the items and could pass null, but if for some reason someone would want to know which item changed the check state, passing that info inside a compound button is problematic (the button might not even exist))

gubatron:
From a more practical point of view, I'm not too comfortable with this:
```
- public void onItemChecked(CompoundButton v, boolean checked) {		 
+public void onItemChecked(T item, boolean checked) {
```

It's always good to have access to the view that triggers the event. If anything, I think I would've just added the parameter, having just the associated model object to the item clicked here doesn't make too much sense to me, I feel like I'd then have to do acrobatics if I wanted to do any UI code

```
public void onItemChecked(CompoundButton v, T item, boolean checked) {
```

this way you avoid doing the `.setTag`, `.getTag` acrobatics

PR #311, opened 2016-12-21
Feature/settings makeover WIP PR

grzesiekrzaca:
as per @gubatron request WIP PR.

### DIFF
--- a/android/src/com/frostwire/android/core/ConfigurationDefaults.java
+++ b/android/src/com/frostwire/android/core/ConfigurationDefaults.java
@@ -102,6 +102,7 @@ final class ConfigurationDefaults {
 
         defaultValues.put(Constants.PREF_KEY_NETWORK_ENABLE_DHT, true);
         defaultValues.put(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA, true);
+        defaultValues.put(Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY, false);
         defaultValues.put(Constants.PREF_KEY_NETWORK_MAX_CONCURRENT_UPLOADS, 3);
 
         defaultValues.put(Constants.PREF_KEY_TORRENT_SEED_FINISHED_TORRENTS, false);
--- a/android/src/com/frostwire/android/gui/activities/SettingsActivity2.java
+++ b/android/src/com/frostwire/android/gui/activities/SettingsActivity2.java
@@ -20,6 +20,7 @@ package com.frostwire.android.gui.activities;
 import android.app.Activity;
 import android.app.Fragment;
 import android.app.FragmentTransaction;
+import android.app.NotificationManager;
 import android.content.Intent;
 import android.os.Bundle;
 import android.preference.CheckBoxPreference;
@@ -31,17 +32,21 @@ import android.preference.PreferenceScreen;
 import android.preference.SwitchPreference;
 import android.preference.TwoStatePreference;
 import android.util.Log;
+import android.view.View;
 
 import com.frostwire.android.AndroidPlatform;
 import com.frostwire.android.R;
 import com.frostwire.android.core.ConfigurationManager;
 import com.frostwire.android.core.Constants;
+import com.frostwire.android.gui.LocalSearchEngine;
 import com.frostwire.android.gui.NetworkManager;
 import com.frostwire.android.gui.SearchEngine;
 import com.frostwire.android.gui.services.Engine;
+import com.frostwire.android.gui.services.EngineService;
 import com.frostwire.android.gui.transfers.TransferManager;
 import com.frostwire.android.gui.util.UIUtils;
 import com.frostwire.android.gui.views.AbstractActivity2;
+import com.frostwire.android.gui.views.preference.ButtonActionPreference;
 import com.frostwire.android.gui.views.preference.CheckBoxSeedingPreference2;
 import com.frostwire.android.gui.views.preference.NumberPickerPreference;
 import com.frostwire.bittorrent.BTEngine;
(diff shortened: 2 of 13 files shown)


aldenml:
Hi @grzesiekrzaca, if you split the changes in smaller PRs, I can speed up the review and merge into master in order to get the rest of the team test it. My recommendation is one PR per single setting/item in the screens.

android/src/com/frostwire/android/core/ConfigurationDefaults.java
```diff
@@ -102,6 +102,7 @@ private void load() {
 
         defaultValues.put(Constants.PREF_KEY_NETWORK_ENABLE_DHT, true);
         defaultValues.put(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA, true);
+        defaultValues.put(Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY, false);
```
gubatron: lets not add another setting, this can already be handled with USE_MOBILE_DATA == false

android/src/com/frostwire/android/gui/activities/SettingsActivity2.java
```diff
@@ -147,6 +309,82 @@ public void onCreate(Bundle savedInstanceState) {
...
+                        activeSearchEnginePreferences.put(preference, engine);
+                    } else {
+                        inactiveSearchEnginePreferences.put(preference, engine);
+                    }
+                }
+            }
+        }
     }
 
     public static class Torrent extends PreferenceFragment {
```
gubatron: the name of this class is very misleading. It should be more like TorrentPreferenceFragment

android/src/com/frostwire/android/gui/activities/SettingsActivity2.java
```diff
@@ -156,6 +394,103 @@ public void onCreate(Bundle savedInstanceState) {
...
+                    }
+                });
+            }
+
+            if (preferenceSeeding != null && preferenceSeedingWifiOnly != null) {
+                preferenceSeedingWifiOnly.setEnabled(preferenceSeeding.isChecked());
+            }
+        }
+
+        //todo think about adding units
```
gubatron: ha, came to say that we should pass units as enums here

android/src/com/frostwire/android/gui/activities/SettingsActivity2.java
```diff
@@ -156,6 +394,103 @@ public void onCreate(Bundle savedInstanceState) {
...
+                            return e.maxPeers() == newVal;
+                        }
+                        return false;
+                    }
+                });
+                displayNumericalSummaryForPreference(pickerPreference, ConfigurationManager.instance().getLong(key), unlimitedValue, rate);
+            }
+        }
+
+        private void displayNumericalSummaryForPreference(Preference preference, long value, Long unlimitedValue, boolean rate) {
```
gubatron: and the units would also be passed here I suppose

android/src/com/frostwire/android/gui/adapters/TransferListAdapter.java
```diff
@@ -318,9 +318,10 @@ private String populateBittorrentDownloadMenuActions(BittorrentDownload tag, Lis
             } else {
                 boolean wifiIsUp = NetworkManager.instance().isDataWIFIUp();
                 boolean bittorrentOnMobileData = ConfigurationManager.instance().getBoolean(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA);
+                boolean bittorrentOnlyWifi = ConfigurationManager.instance().getBoolean(Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY);
```
gubatron: wipe this out
bittorrentOnlyWifi == !bittorrentOnMobileData

it is redundant here, will just add headaches.

android/src/com/frostwire/android/gui/services/EngineBroadcastReceiver.java
```diff
@@ -122,6 +122,7 @@ private void handleConnectedNetwork(NetworkInfo networkInfo) {
         if (NetworkManager.instance().isDataUp()) {
 
             boolean useTorrentsOnMobileData = ConfigurationManager.instance().getBoolean(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA);
+            boolean bittorrentOnlyWifi = ConfigurationManager.instance().getBoolean(Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY);
```
gubatron: wipe out

android/src/com/frostwire/android/gui/transfers/TransferManager.java
```diff
@@ -411,7 +411,8 @@ public boolean isBittorrentDownload(Transfer transfer) {
 
     public boolean isMobileAndDataSavingsOn() {
         return NetworkManager.instance().isDataMobileUp() &&
-                !ConfigurationManager.instance().getBoolean(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA);
+                (!ConfigurationManager.instance().getBoolean(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA)
```
gubatron: same

android/src/com/frostwire/android/gui/views/preference/CheckBoxSeedingPreference2.java
```diff
@@ -0,0 +1,81 @@
...
+import android.view.ViewGroup;
+import android.widget.CheckBox;
+import android.widget.TextView;
+
+import com.frostwire.android.R;
+
+/**
+ * @author gubatron
+ * @author aldenml
+ */
```
gubatron: @author grzesiek :)

android/src/com/frostwire/android/gui/activities/SettingsActivity2.java
aldenml: I don't think it's misleading, that class is inside `SettingsActivity2`, and that activity should only expose fragments to be assembled in a final preference screen xml. In this case, it's like having enums suffixed by `_ENUM`.

aldenml:
I want to add that this is an excellent opportunity to revisit old decisions and simplify code. Let's do this in a granular way. I understand this is a WIP PR, but some of the things here can be easily merged in isolation.

android/src/com/frostwire/android/gui/activities/SettingsActivity2.java
gubatron: shouldnt it be private static then?

im just thinking of code completion and how you can end up having this on a completey unrelated context. This is not a Torrent, if anything a TorrentPreference

android/src/com/frostwire/android/gui/adapters/TransferListAdapter.java
gubatron: or i guess another way to go about this, is that once the setting screen is merged you can rename the older boolean setting to reflect the new ui (wifi_only) and then negate EVERY usage. I think this would be an unnecessary risk to introduce bugs

aldenml: a lot of "unnecessary risk"s

grzesiekrzaca:
@aldenml 
Ok tomorrow I'll split it into more PRs

@gubatron 
I added the new preference because the old one was the negation of this new one and that affects how it looks as a view (preference is true - switch is "on" so we cannot use the old one 'as is') if I wanted to keep the old one I would have to negate it everywhere in the code - that could lead to a) bugs if I missed it b) confusion in the future (the name of preference is `USE_MOBILE_DATA` but it uses mobile data when it is false). [not to mention sudden change for every user that updates the app] The other problem is coexistence in the current code base. As it is (in my code) an AND on two preferences one of which is set to true (negated false) nothing is changed if someone uses the old settings menu.

gubatron:
>(the name of preference is USE_MOBILE_DATA but it uses mobile data when it is false)

Only for cloud downloads, shouldn't do so for BitTorrent, as specified in the old settings summary.

But yes, we decided now, that we will:
1. Stop all transfers when switching to USE_MOBILE_DATA = False and when we are on a non-wifi connection.
2. Not start a cloud transfer when same conditions apply.

Now, if you truly want to keep the setting that reflects the way the setting looks like in the UI, please remove all usages of the old setting. It's always a bad idea to have 2 boolean variables that negate each other and control the same thing.

android/src/com/frostwire/android/gui/views/preference/CheckBoxSeedingPreference2.java
```diff
@@ -0,0 +1,81 @@
...
+public class CheckBoxSeedingPreference2 extends android.preference.CheckBoxPreference {
```
gubatron: I suppose this one will be replaced when we merge your new PR

grzesiekrzaca: Yes, this branch is stale.

grzesiekrzaca:
super stale branch

PR #317, opened 2016-12-28
Feature/granular/torrent preferences sharing

grzesiekrzaca:


aldenml:
Question: What is the usability use case for `CheckBoxSeedingPreference2`? Is it something specified in another ticket?

android/src/com/frostwire/android/gui/views/preference/CheckBoxSeedingPreference2.java
```diff
@@ -0,0 +1,134 @@
...
+ *         notifyChanged() in Preference breakes the animation of the checkbox.
+ *         Preference logic has to be rewritten not to use it. (Manual manipulation of views)
+ */
+public class CheckBoxSeedingPreference2 extends android.preference.CheckBoxPreference {
+    private final ColorStateList titleColor;
+    private final ColorStateList summaryColor;
+    private TextView title;
+    private TextView summary;
+    private CheckBox checkbox;
+    private Boolean checked;
```
aldenml: Does this needs to be a class?

grzesiekrzaca:
It was not specified in a ticket per se but this allows a different disabled state than normal preference. 
Same as `CheckBoxSeedingPreference` (code in `SettingsActivity2` related to this is a copy-paste) except @marcelinkaaa wanted that to be animated (the old one was not due to way `setChecked()` worked) [also I saw that the title and summary text was of the same color and changed that to have different colors.]

android/res/color/preference_summary_text.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gubatron: don't forget to add the license headers, and also include yourself as one of the authors.

grzesiekrzaca: hmm, I am confused as to which one. I see that some files are with GNU and some are with Apache. The readme states that it should be GNU, but your cleanup PR on my branch had the Apache.

gubatron: At some point we started adding Apache licenses, I can't remember if we were continuing to change to Apache.

Files we have written from scratch should have no issue becoming Apache, however, files from other projects we can't change the header.

As an author, I'd leave it up to you to decide.

I think Apache suits best for things like utility classes, libraries (like jlibtorrent, but we license that one under MIT), but things that are very specific to FrostWire that some other competing client could copy we should probably leave as GNU to keep the code free.

I'm also confused to be honest, @aldenml knows better what was that decision.

aldenml:
Is it possible to put the changes of `CheckBoxSeedingPreference2` in the old one?

grzesiekrzaca:
@aldenml I tried to (just now) but it breaks padding in the old `SettingsActrivity` (not really sure why). Should I continue to find a solution or just leave 'as is' and proceed with the old plan that once we have everything working we remove old classes and rename those with 2s. (Same thing with #318 `ToggleAllSearchEnginesPreference2`)

grzesiekrzaca:
This needs #323 to be merged first

aldenml:
I'm not sure `CheckBoxSeedingPreference2` provide any advantage (or even a good feature) here. I rather remove it from this PR to get this merged and we can discuss the "multicolor" feature later.

grzesiekrzaca:
Now that there is no need to do weird stuff relating to the lifecycle of the preference we could change that to a custom layout (see   https://github.com/frostwire/frostwire/pull/317/commits/a8287aaba98696b6feb6396271a986745097bec8 )
I think that graying out the text as well as the checkbox looks good. @marcelinkaaa what do you think?

marcelinkaaa:
Yes graying out works perfectly. And animations look good as well. I still have to go over the code to see exactly what was done but on the surface works fine. 
We should, however, review the summary on 'Seed Finished Torrents' and 'Seed only on Wi-Fi'. There should be only one (no matter if its checked or not) on the first one and the Wi-Fi one is inaccurate.

android/res/layout/view_preference_greyed_out_on_disable.xml
```diff
@@ -0,0 +1,64 @@
...
+    </RelativeLayout>
+
+    <!-- Preference should place its actual preference widget here. -->
+    <LinearLayout android:id="@android:id/widget_frame"
+        android:layout_width="wrap_content"
+        android:layout_height="match_parent"
+        android:gravity="end|center_vertical"
+        android:paddingLeft="16dp"
+        android:orientation="vertical" />
+</LinearLayout>
```
aldenml: \n

PR #318, opened 2017-01-02
Feature/granular/search properties

grzesiekrzaca:
search engines list with animated check all


aldenml:
I was expecting to get rid of `ToggleAllSearchEnginesPreference2`

android/src/com/frostwire/android/gui/activities/ToggleAllSearchEnginesPreference2.java
```diff
@@ -0,0 +1,205 @@
...
+ * @author grzesiekrzaca
+ */
+public class ToggleAllSearchEnginesPreference2 extends CheckBoxPreference {
+
+    private CheckBox checkbox;
+    private Map<CheckBoxPreference, SearchEngine> searchEnginePreferences;
+    private int backgroundColor;
+    private boolean fixStateAfterNextBind = true;
+
+    private View.OnClickListener checkBoxListener = new View.OnClickListener() {
```
gubatron: why keep a reference to this `checkBoxListener` if it's not called anywhere else outside `onCreateView()`.

another way would be to have this class implement View.OnClickListener, and then the `onClick(View v)` implementation calls `onClick()`.

Let's try not to keep unnecessary properties and keep scopes as local as possible.

android/src/com/frostwire/android/gui/activities/ToggleAllSearchEnginesPreference2.java
```diff
@@ -0,0 +1,205 @@
...
+                } else if (isOneEngineNotChecked()) {//turnoff the check (just unchecked on one of the children)
+                    checkbox.getHandler().post(delayedCheckboxCheck);
+                }
+            } else {
+                checkbox.setChecked(isChecked());
+            }
+            return true;
+        } //can't change the child, don't change anything
+        return false;
+    }
```
gubatron: all these checks seem to eventually do the same thing,

If the checkbox has a handler do `checkbox.setChecked(isChecked());` now.

If all engines are checked, do `checkbox.setChecked(isChecked());` later (aka `delayedCheckboxCheck`)

if one of the engines isn't checked do `checkbox.setChecked(isChecked());` later

Therefore, it seems to me that the `isOneEngineNotChecked()` method isn't needed because you will perform the check anyway.

Perhaps this could be as simple as:

```java
...
if (checkbox.getHandler() != null) {
  checkbox.getHandler().post(delayedCheckboxCheck);
} else {
  checkbox.setChecked(isChecked());
}
return true;
...
```

I'd also rename `isAtLeastOneOtherEngineChecked` to `isOneEngineChecked()`

and would reimplement the whole thing like

```java
public boolean requestChildStateChange(CheckBoxPreference checkBoxPreference) {
    if (!isOneEngineChecked(checkBoxPreference)) {
        // we're done here
        return false;
    }
    if (checkbox.getHandler() != null) {
       // check later
        checkbox.getHandler().post(delayedCheckboxCheck);
    } else {
       // check now
       checkbox.setChecked(isChecked());
    }
    return true;
}
```

Then you can get rid of that `isOneEngineNotChecked()` method and I think it's a lot easier to read.

However, I'd try to re-read the whole thing and see how to simplify, it seems like there are a lot of hacks that will make this hard to maintain. Like what happens if this method returns `true` when invoked above:

```
 if (!selectAll.requestChildStateChange((CheckBoxPreference) cbPreference)) {
                         cbPreference.setChecked(!cbPreference.isChecked());//reverting change we need one
                     }
```

then there's also what happens onBind, not sure if that's there to fix a previous bug, but in my experience, when you start having code that looks like that, at some point later in the future it ends up being removed because it just looks unnecessary.

grzesiekrzaca:
@gubatron 
Thanks. That is better.
I had a lot of problems when dealing with this class and different ways of fixing them. Those are echoes of those problems.
The onBind 'fix' is needed. When we create the preference we don't have the list of children so the isChecked() shows false. We need to fix that later. (comment this code, check all, scroll to bottom uncheck one of the preferences, press back, press Search - wrong state)
I chose another name for the function (and inverted the return) I think that better reflects what we are checking in it. 
`SettingsActivity2` - `onPreferenceClick` was pointless, changed to `onPreferenceChange` and used that function return "mechanism".

android/src/com/frostwire/android/gui/activities/ToggleAllSearchEnginesPreference2.java
```diff
@@ -152,32 +144,15 @@ private boolean areAllEnginesChecked() {
...
         for (CheckBoxPreference preference : searchEnginePreferences.keySet()) {
             if (preference != null && preference != checkBoxPreference) {
                 if (preference.isChecked()) {
-                    return true;
+                    return false;
                 }
             }
         }
-        return false;
+        return true;
```
gubatron: I see the name suggests that you're verifying that this checkbox preference is the only one that has been checked.

What I see here, is that you return false if some other is checked, but at the end of the method, you don't check if the given checkbox has been checked, perhaps it's not and this is returning true.

gubatron:
good to see that it's been simplified a lot. Left a comment on one of the methods, perhaps the name threw me off, perhaps it's missing a check.

Thanks for the explanation of the onBind

gubatron:
I will now need to fetch this branch and test it to give you my ACK, it's a lot better now :)

android/src/com/frostwire/android/gui/activities/ToggleAllSearchEnginesPreference2.java
grzesiekrzaca: You are right. This code will "break" if we have nothing checked (will not happen, but still...) we would not allow checking the preference. I added a check.

aldenml:
I always thought that `ToggleAllSearchEnginesPreference2` (including the internal hacks) was unnecessary, see https://github.com/frostwire/frostwire/commit/c29ba4a35615b3a6a72da7d233386078f2e43ab4

Feel free to close this PR unless you detect something is missing in master

grzesiekrzaca:
@aldenml Looks great, closing.

PR #319, opened 2017-01-02
[android] copy other preferences

grzesiekrzaca:
This is a copy of existing code with the exception of the comment about UXStats. I think that that should be tackled in another PR as it would touch update logic.

### DIFF
--- a/android/res/xml/settings_application.xml
+++ b/android/res/xml/settings_application.xml
@@ -53,7 +53,7 @@
 
     <android.support.v7.preference.PreferenceCategory android:title="@string/other_settings">
         <android.support.v7.preference.Preference
-            android:fragment="com.frostwire.android.gui.activities.SettingsActivity2$Other"
+            android:fragment="com.frostwire.android.gui.fragments.preference.OtherFragment"
             android:title="@string/notification_other_header" />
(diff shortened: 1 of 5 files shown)


aldenml:
Why so many things imports?

grzesiekrzaca:
Forgot to `Ctrl+Alt+O` after cherry picking WIP branch commits

gubatron:
@aldenml when you come back, let's make it a priority helping @grzesiekrzaca get all the Settings work  finished and merged for a release. Still a lot of work to be done after this on My Files (Material Design, New Push and Hold Selection Modes, and grid display of images and videos), of course always adhering to your good standards.

Also let's make sure once this happens, that we don't have a single "SomeClass2" in the code base, that everything is finally replaced.

I'll be at CES until sunday, I'll try to connect and work for half of the day.

aldenml:
I took over this because `R.layout.view_preference_button_action` wasn't the correct layout to use and there was a good opportunity to simplify `ButtonActionPreference2`. See https://github.com/frostwire/frostwire/commit/edaa55cac73e1a25b4b7d67064db3815f41c1d9d

Feel free to close this PR.

PR #323, opened 2017-01-10
Migration to support preferences

grzesiekrzaca:
`CheckboxSeedingPreference` was commented out as it is still used in `SettingsActivity` and there is another PR that has a new class for that Preference and just needs to be updated to use support preferences


aldenml:
What is the problem with dialog icon and icon?

I think it is better if we follow the pattern like in this post: https://medium.com/@JakobUlbrich/building-a-settings-screen-for-android-part-3-ae9793fd31ec#.p6t2d37np

grzesiekrzaca:
@aldenml 
I'm using the newer versions of support libraries. *Compat stuff is from v7 and makes us use the v4 Fragments. That, in turn, would change SettingsActivity2 to also use those Fragments. I think it's cleaner to use normal Fragments.

As to the icon when only null the LinearLayout is still there (see the comment above the code that hides it):
screenshot_1

aldenml:
I think it's time to move `SettingsActivity2.Torrent` to a new file/class `com.frostwire.android.gui.fragments.preference.TorrentFragment extends AbstractPreferenceFragment`

grzesiekrzaca:
@aldenml 
I was just writing a slack message about that :)

android/src/com/frostwire/android/gui/views/preference/NumberPickerPreference2.java
```diff
@@ -0,0 +1,148 @@
...
+    private int endRange;
+    private int defaultValue;
+
+    public NumberPickerPreference2(Context context, AttributeSet attrs) {
+        super(context, attrs);
+
+        if (attrs == null) {
+            return;
+        }
+        setIcon(null);
```
aldenml: is this necessary? if not, it can be removed

android/src/com/frostwire/android/gui/fragments/preference/TorrentFragment.java
```diff
@@ -0,0 +1,164 @@
...
+import com.frostwire.android.R;
+import com.frostwire.android.core.ConfigurationManager;
+import com.frostwire.android.core.Constants;
+import com.frostwire.android.gui.util.UIUtils;
+import com.frostwire.android.gui.views.AbstractPreferenceFragment;
+import com.frostwire.android.gui.views.preference.NumberPickerPreference2;
+import com.frostwire.bittorrent.BTEngine;
+
+
+public class TorrentFragment extends AbstractPreferenceFragment {
```
aldenml: can you make it final: `public final class`?

android/src/com/frostwire/android/gui/views/preference/NumberPickerPreference2.java
```diff
@@ -0,0 +1,148 @@
...
+import android.widget.NumberPicker;
+
+import com.frostwire.android.R;
+
+/**
+ * @author grzesiekrzaca
+ *
+ * Support version of a custom dialog preference
+ */
+public class NumberPickerPreference2 extends DialogPreference {
```
aldenml: `final`

android/src/com/frostwire/android/gui/views/preference/NumberPickerPreference2.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    public int getDefaultValue() {
+        return defaultValue;
+    }
+
+    /**
+     * Actual dialog used to interact with the preference
+     */
+    public static class NumberPickerPreferenceDialog extends PreferenceDialogFragment {
```
aldenml: final

android/src/com/frostwire/android/gui/views/preference/NumberPickerPreference2.java
```diff
@@ -0,0 +1,148 @@
...
+            Bundle bundle = new Bundle(4);
+            bundle.putString(ARG_KEY, preference.getKey());
+            bundle.putInt(START_RANGE, ((NumberPickerPreference2) preference).getStartRange());
+            bundle.putInt(END_RANGE, ((NumberPickerPreference2) preference).getEndRange());
+            bundle.putInt(DEFAULT_VALUE, ((NumberPickerPreference2) preference).getDefaultValue());
+            fragment.setArguments(bundle);
+            return fragment;
+        }
+    }
+}
```
aldenml: new line '\n' missing

gubatron:
yay

PR #327, opened 2017-01-18
Material/styled toasts

grzesiekrzaca:
Making toasts in Settings look like in the mockups


android/src/com/frostwire/android/gui/fragments/preference/OtherFragment.java
```diff
@@ -110,7 +110,7 @@ private void setupClearIndex() {
                 @Override
                 public void onClick(View v) {
                     LocalSearchEngine.instance().clearCache();
-                    UIUtils.showShortMessage(getActivity(), R.string.deleted_crawl_cache);
```
gubatron: Why don't we just make the old `UIUtils.showShortMessage` use the material style if it can, otherwise fall back to non styled toasts?

CC: @marcelinkaaa, I guess I just want to know if you want to now style all toasts, if so, let's do what I said above, and avoid having 2 methods for the same purpose.

marcelinkaaa: @gubatron the most important thing for me is to show the two toast messages in settings now. The "By default, one search engine needs to be checked at all times" and hopefully "Disconnected from BitTorrent. All BitTorrent transfers have stopped." 

I'm not sure if you can only use material style leaving everything else as is Gubatron (I would style all toasts now, btw). But it also seems like a lot additions. Is some java without adding layouts possible? ex. http://www.androidhive.info/2015/09/android-material-design-snackbar-example/ @grzesiekrzaca

gubatron: ok, so it's a compatibility issue? meaning, those toasts need something loaded before hand?
I'd think toasts would be sort of independent windows.
I'll have to test the branch myself tomorrow.
If I make it work, I'll submit a PR to greg on this branch.

aldenml:
I do think this is the wrong approach, why not use the `Snackbar` instead?

gubatron:
#TIL about the new "Snackbar". What do you think about it @marcelinkaaa

grzesiekrzaca:
I implemented the Snackbar but left the toast as a fallback.

gubatron:
way to go!, thanks for the fast iteration :)

aldenml:
Why do you need to use `CoordinatorLayout` and `AppBarLayout`?
Why do you keep `showMaterialToastMessage` as a fallback?

There is a reason `Snackbar` works with `View` and `Toast` with `Context`.

grzesiekrzaca:
`CoordinatorLayout` allows us to dismiss `Snackbar` with swipe
`AppBarLayout` is needed in `CoordinatorLayout` to make `Toolbar` to be displayed correctly.

I thought to have a fallback if the new Material message would be opened from somewhere that has no `CoordinatorLayout`. Should I delete it?

aldenml:
If the proposed message is short intentionally, there should be no need of swipe-to-dismiss. If in the future we want to provide this feature universally, a consistent design across all activities must be in place (and done in another PR). I would say that anything that it's not used, should not be added in the first place.

I expect this PR collapse to a few lines.

android/res/values/colors.xml
```diff
@@ -78,4 +78,6 @@
     <color name="warning_red">#ffff0000</color>
     <color name="approval_green">#05a751</color>
 
+    <color name="toast_background">#ff333333</color>
```
aldenml: remove

android/src/com/frostwire/android/gui/util/UIUtils.java
```diff
@@ -113,6 +114,10 @@ private static void showToastMessage(Context context, String message, int durati
         }
     }
 
+    public static void showShortMaterialMessage(View view, String message) {
```
aldenml: why not call this `showShortMessage`, and if you receive a `resId`, you don't need to call `getString` while using it

PR #330, opened 2017-01-23
Material/custom pickers

grzesiekrzaca:


aldenml:
What's this?

grzesiekrzaca:
Better pickers for numbers. Using a spinner is a PITA for anything large.

aldenml:
And what happens for smaller ranges? (and I'm pretty sure `PreferenceDialogFragmentCompat` is not the way to go)

grzesiekrzaca:
Same thing.  @marcelinkaaa Do we want spinners back for small ranges?
I don't like `PDFC` but we need the support version of `AlertDialog`.

marcelinkaaa:
@grzesiekrzaca As to the font - because dialogs repeat I will add medium in styles for 21 and up. Bold will the fallback. And, yes set it as main for now. 

@aldenml @gubatron For smaller ranges I do like pickers but in our case I would prefer to keep the same dialogs; the default will be set and with a clear description of values to choose from including the addition of reset button I think the dialogs that Grzesiek is working on will work just fine or even better. 
Don't know of any study or don't know anyone who tested which dialogs work best for the users, so can not tell if one is that much better then other in this case.

aldenml:
and why do you need to work directly with `AlertDialog`?

grzesiekrzaca:
To show it. If we use the original `AlertDialog` we get wrong layouts. For some reason, setting `android:alertDialogStyle` or `android:alertDialogTheme` in the Material theme does not change the layout of the dialog.
If on the other hand, we use support version of `AlertDialog` it gives the correct layout. 
The problem is that the `PreferenceDialogFragment` uses original `AlertDialog` and I cannot just extend it and change the on Create as it is tied to the whole lifecycle (saving the values to saved instance state).

aldenml:
I strongly believe using `AlertDialog` is not the way to go. Take a look at `NumberPickerPreference2` and `KitKatStoragePreference`

grzesiekrzaca:
`KitKatStoragePreference` uses `KitKatStoragePreferenceDialog` which uses an `AlertDialog` because It (`KSPD`) extends `PreferenceDialogFragment` which uses `AlertDialog`.
Both screenshots on API 16:  
`KitKatStoragePreferenceDialog` right now (not Material): 
876952293432c5599f129293796f2b0b
`KitKatStoragePreferenceDialog` on this branch (wrong padding in the list ine the background - I'll rebase in a sec) with the change of `PreferenceDialogFragment` to `PreferenceDialogFragmentCompat` (Material but some padding in custom layout is needed)
27b2cbdfd1f05eb8917384d0ba20640b

aldenml:
After researching about it, I came to the same conclusion that v7 `AlertDialog` is needed. However, extending (with minor reflection tricks) is possible. See https://github.com/frostwire/frostwire/commit/578cb63400f0ea9de074f794d558625b7f157b6d#diff-70b24681a12db6af1d69ccad99e0c3c3R83 (you can rebase and use it)

@marcelinkaaa some minor styles modifications are needed, can you take a look?

grzesiekrzaca:
I'm paranoid about reflection but if you think that is the way I'll analyze that approach to get `AlertDialog` in correct form

gubatron:
testing this branch together @marcelinkaaa today

aldenml:
@gubatron this still far from being in a good shape code wise

gubatron:
the other (older) picker allowed for both manually entering an amount and to scroll up and down.
However, the intervals shown before (1 by 1 byte) were unusable.

@marcelinkaaa says up to this day she had no clue she could touch the number and enter a number manually. Perhaps what I'll do is add a hint close to the title, letting the user know they can enter a number manually. (this will most likely be a hack, we'll see)

I'm going to work on this, I really want to release this week, this is getting ridiculously long.

gubatron:
>Better pickers for numbers. Using a spinner is a PITA for anything large.

try touching the number at the center of the picker. that picker is cool because it allows for both manual and spinning, the only problem is that the spinner is configured at the wrong interval.

I'll try and fix this on master

gubatron:
update:

Android's number picker is a piece of shit.

Reasons:
-> There's no built-in "step" parameter.
-> They have a way to let you enter an array of values to display, I finally got to hack it all the way and show a configurable number of parameters, but the issue then is that when you try to use the keyboard if you want to enter a custom value, it will try to autocomplete to one of the pre-loaded values.

it's a control that's useless for this case.

I will try using a seekbar to see how it behaves, if this doesn't work, we'll have to stick to the patch that greg sent on which the Max Download Speed and Max Upload Speed dialogs work with a text entry, in which case I'd add the possibility of switching the unit, so if the user wants to enter megabytes or gigabytes they don't have to count all the zeroes.

aldenml:
Well, there is no reason to not use the current one for "max downloads" and "max uploads". If there is a need for a new control, it should be only for the needed ones.

Then, I see greg is still dealing with `AlertDialog` in a few places, I don't think that's necessary (or even a good idea). The input validation seems to me a little unsatisfactory, free text input (without a hard mask) is not suitable for numeric inputs in a small device.

And what about the dealing with the styles, @marcelinkaaa can you take a look at it?

marcelinkaaa:
@aldenml @gubatron re dealing with styles. Well I am still not a 100% sure if we are leaving the current picker after today. For high numbers as we are using now for speeds I definitely am recommending a text input very much alike to what @grzesiekrzaca is working on. 

If we leave the number picker - I know the text size of the numbers was much smaller. I looked into it and it looks like another Android bug. It displays well on API 16-22, then 23 up the sizes are lost. Unfortunately everything I found on the subject today required quite a bit of hacking, styling is very complicated here - as there is no direct way to style the textSize. What seem to work pretty well for now is simply adding scaleX=1.3 and scaleY=1.3 to number picker in the layout. I also want to try one more thing in the morning and will let you know. 

But please do let me know what you guys decide to do first. Are we sticking with what we have, trying the seekbar (which I rather not have controlling high numbers), or going with the text input.

gubatron:
This is waaaaaay better. Almost done. Sucks I lost all day because it would've worked real nice with the number picker had the number picker been more capable, but this is way better.

screen shot 2017-01-26 at 10 05 15 pm

and if you go to zero, it displays itself as Unlimited
screen shot 2017-01-26 at 10 06 49 pm

gubatron:
since I use the function that displays bytes in human friendly units, also the API will be simplified more, and it's just super friendly to the user, it will jump from bytes to kbs, to mbs as you move the slider.

gubatron:
Adjusting max values to 25Mb, then you get KB granularity
screen shot 2017-01-26 at 10 12 43 pm

screen shot 2017-01-26 at 10 12 52 pm

screen shot 2017-01-26 at 10 13 00 pm

screen shot 2017-01-26 at 10 13 04 pm

I find this to be waaaaaaaay better to use, I hope you can agree when you try it.

Perhaps what I will do will be to hack it so that if you go all the way to the right, then that means "Unlimited", as `0` is antinatural to a regular person as unlimited.

aldenml:
it looks good, will be this another preference type? is it custom or from the framework? will be it used even for smaller ranges (like max uploads)?

grzesiekrzaca:
@aldenml 
I don't understand the problem with validation.
`EditText` has:
`android:inputType="number"`
so one cannot enter any text
Also, I'm touching `AlertDialog` for two reasons: 
1) disable set while out of range value is entered 
2) change neutral button behavior so it doesn't close the dialog

@gubatron 
I don't know how to proceed with this.

marcelinkaaa:
@gubatron well, let me know when I can test. I'll see if anything needs to be adjusted across APIs, looks good style wise on screenshots.

gubatron:
>it looks good, will be this another preference type? is it custom or from the framework? 

I'm thinking of naming it `SeekBarPreference`, perhaps there's one already like that, and yes it will be custom as it's richer and takes into consideration byteRates vs cardinals.

gubatron:
I'll be submitting/merging a PR this afternoon with my finished branch, dealing with all the torrent preferences. @grzesiekrzaca 

Right now I'm running medical errands, should be back at 12:45pm my time.

PR #334, opened 2017-01-24
Feature/delete torrent on download

grzesiekrzaca:
Adding back for review

### DIFF
--- a/common/src/main/java/com/frostwire/bittorrent/BTEngine.java
+++ b/common/src/main/java/com/frostwire/bittorrent/BTEngine.java
@@ -330,6 +330,18 @@ public final class BTEngine extends SessionManager {
         }
     }
 
+    public void deleteTorrentFile(TorrentInfo ti) {
+        entry entry = ti.toEntry().swig();
+        File torrentFile = new File(entry.dict().get(TORRENT_ORIG_PATH_KEY).string());
+        try {
+            if (!torrentFile.delete()) {
+                LOG.warn("Error deleting torrent file");
+            }
+        } catch (Exception e) {
+            LOG.warn("Error deleting torrent file ", e);
+        }
+    }
+
     public void restoreDownloads() {
         if (swig() == null) {
             return;
--- a/common/src/main/java/com/frostwire/bittorrent/BTEngine.java
+++ b/common/src/main/java/com/frostwire/bittorrent/BTEngine.java
@@ -452,10 +452,9 @@ public final class BTEngine extends SessionManager {
         try {
             String name = getEscapedFilename(ti);
 
-            byte[] arr = ti.toEntry().bencode();
-
             entry e = ti.toEntry().swig();
             e.dict().set(TORRENT_ORIG_PATH_KEY, new entry(torrentFile(name).getAbsolutePath()));
+            byte[] arr = Vectors.byte_vector2bytes(e.bencode());
 
             FileUtils.writeByteArrayToFile(resumeTorrentFile(ti.infoHash().toString()), arr);
         } catch (Throwable e) {
--- a/common/src/main/java/com/frostwire/bittorrent/BTEngine.java
+++ b/common/src/main/java/com/frostwire/bittorrent/BTEngine.java
@@ -261,7 +261,7 @@ public final class BTEngine extends SessionManager {
         download(ti, saveDir, selection, peers, false);
     }
 
-    public void download(TorrentInfo ti, File saveDir, boolean[] selection, List<TcpEndpoint> peers, boolean deleteTorrentFile) {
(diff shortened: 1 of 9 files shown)


aldenml:
I don't quite understand the logic behind this feature. Torrents are saved for convenience, but anyway, I think that if a user doesn't want to have the torrent saved, why save it in the first place? Saving and deleting it looks weird at best.

grzesiekrzaca:
I changed the logic in `BTEngine`. Now we just don't save the file, only resume file. 
@marcelinkaaa maybe this should be a reversed setting? Ie. Keep downloaded .torrent files? default to true?

gubatron:
@aldenml main motivation is privacy. second motivation is that other clients have this and it's becoming expected to have it. some users might not want to keep torrents and it's an extra step they have to take every single time, an excuse to not use FrostWire.

@grzesiekrzaca I prefer the affirmative setting defaulted to false just as specified in the Issue request. This is how users will expect it as seen in other clients.

common/src/main/java/com/frostwire/bittorrent/BTEngine.java
```diff
@@ -258,6 +258,10 @@ public void download(File torrent, File saveDir, boolean[] selection) {
     }
 
     public void download(TorrentInfo ti, File saveDir, boolean[] selection, List<TcpEndpoint> peers) {
+        download(ti, saveDir, selection, peers, false);
```
gubatron: shouldn't this parameter be the current config value instead of defaulting to `false`?

gubatron: oh sorry, this is `common`

aldenml:
privacy wise, this does not help, but if other clients have it, it's good we replicate it too. But my main concern was the save/delete logic, instead of just not saving. I see it's changed now, but need to review the code

marcelinkaaa:
@grzesiekrzaca we are a torrent downloader first so lets not reverse the setting. @gubatron I actually do like the addition but didnt find the setting in many apps yet. Shutdown app after download completed and set a port number were the top two I think.

gubatron:
so I was reading the code and I believe I understand why the save/delete logic here:

```java
if (!exists) {
    File torrent = saveTorrent(ti);
    saveResumeTorrent(torrent); // <- I think this is the issue.
    if (deleteTorrentFile) {
        deleteTorrentFile(torrent);
    }
}
```

If I understand correctly, `saveTorrent(TorrentInfo ti)` saves on the user facing `Torrents` folder and `saveResumeTorrent(File torrent)` saves the `.torrent` inside jlibtorrent's inner folder. 

I believe the problem is that this method takes a `File` object, and probably to minimize changes he decided to save the file and delete it.

If we made `saveResumeTorrent()` receive a `TorrentInfo` object, then the code could be simplified to:

```java
saveResumeTorrent(ti);
if (!deleteTorrentFile) {
    saveTorrent(ti);
}
```

and we get rid of the deletion method.

grzesiekrzaca:
@gubatron 
You were right I wanted to minimize change due to the proximity of library code. I changed the code (perhaps you were looking at the specific commit). I'll see if I can change stuff to be closer to what you show.

gubatron:
LGTM now

aldenml:
@gubatron I still have some concerns related to how the feature is worded in the setting

gubatron:
I also remembered that this could probably break desktop and we should check if the desktop build isn't affected.

Will check wording once again, (shouldn't be a show-stopper for the merge), however, breaking desktop build could be.

@grzesiekrzaca have you ever built the desktop client? let me know if you need help.

It's probably a good time to see if the instructions on `desktop/README.md` are up to date.

To run in IntelliJ your run configuration should look like this (so jlibtorrent and other native libraries work properly):
screen shot 2017-01-31 at 8 22 00 am

grzesiekrzaca:
I was able to successfully build the desktop app and run it. (Just had to point it to 32-bit JRE due to jlibtorrent.dll in lib/native)

aldenml:
Well, unfortunately, this is not right at all. In `master`, when you save the torrent file, it's saved as is whenever is possible, but with this change, you are saving a "file representation" of the `TorrentInfo` that in many ways it could be different from the original "torrent file".

grzesiekrzaca:
Unless I don't understand something, this produces same files as `master` for torrents. Resume torrents had a problem that the original file (@key `TORRENT_ORIG_PATH_KEY`) was not saved. I changed that in this commit.

aldenml:
I believe now you are right, I got confused at some detail. Maybe this entire `BTEngine` could (or should) be refactored for simplification.

gubatron:
party

PR #335, opened 2017-01-24
[android] select all ux changes

grzesiekrzaca:
resubmitted for review

### DIFF
--- a/android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
+++ b/android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
@@ -17,14 +17,22 @@
 
 package com.frostwire.android.gui.fragments.preference;
 
+import android.app.Activity;
 import android.support.v7.preference.CheckBoxPreference;
 import android.support.v7.preference.Preference;
+import android.support.v7.preference.PreferenceGroup;
+import android.support.v7.preference.PreferenceGroupAdapter;
+import android.support.v7.preference.PreferenceScreen;
+import android.support.v7.preference.PreferenceViewHolder;
+import android.support.v7.widget.RecyclerView;
 
 import com.frostwire.android.R;
 import com.frostwire.android.gui.SearchEngine;
 import com.frostwire.android.gui.util.UIUtils;
 import com.frostwire.android.gui.views.AbstractPreferenceFragment;
+import com.frostwire.util.Ref;
 
+import java.lang.ref.WeakReference;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -33,6 +41,8 @@ import java.util.Map;
  * @author aldenml
  */
 public final class SearchFragment extends AbstractPreferenceFragment {
+
+    public static final String PREF_KEY_SEARCH_SELECT_ALL = "frostwire.prefs.search.preference_category.select_all";
 
     private final Map<CheckBoxPreference, SearchEngine> activeSearchEnginePreferences;
 
@@ -47,7 +57,7 @@ public final class SearchFragment extends AbstractPreferenceFragment {
     }
 
     private void setupSearchEngines() {
-        final CheckBoxPreference selectAll = findPreference("frostwire.prefs.search.preference_category.select_all");
+        final CheckBoxPreference selectAll = findPreference(PREF_KEY_SEARCH_SELECT_ALL);
 
         Map<CheckBoxPreference, SearchEngine> inactiveSearchPreferences = new HashMap<>();
         fillSearchEnginePreferences(activeSearchEnginePreferences, inactiveSearchPreferences);
@@ -64,6 +74,7 @@ public final class SearchFragment extends AbstractPreferenceFragment {
                         cb.setChecked(true); // always keep one checked
                         UIUtils.showShortMessage(getView(), R.string.search_preferences_one_engine_checked_always);
                     }
+                    selectAll.setTitle(R.string.select_all);
                 } else {
                     updateSelectAllCheckBox();
                 }
@@ -83,7 +94,9 @@ public final class SearchFragment extends AbstractPreferenceFragment {
         selectAll.setOnPreferenceClickListener(new Preference.OnPreferenceClickListener() {
             @Override
             public boolean onPreferenceClick(Preference preference) {
-                checkAllEngines(((CheckBoxPreference) preference).isChecked());
+                CheckBoxPreference selectAll = (CheckBoxPreference) preference;
+                checkAllEngines(selectAll.isChecked());
+                selectAll.setTitle(selectAll.isChecked() ? R.string.deselect_all : R.string.select_all);
                 return true;
             }
         });
@@ -91,9 +104,10 @@ public final class SearchFragment extends AbstractPreferenceFragment {
     }
 
     private void updateSelectAllCheckBox() {
-        CheckBoxPreference cb = findPreference("frostwire.prefs.search.preference_category.select_all");
+        CheckBoxPreference cb = findPreference(PREF_KEY_SEARCH_SELECT_ALL);
         boolean allChecked = areAllEnginesChecked(activeSearchEnginePreferences, true);
         setChecked(cb, allChecked, false);
+        cb.setTitle(allChecked ? R.string.deselect_all : R.string.select_all);
     }
 
     private void fillSearchEnginePreferences(Map<CheckBoxPreference, SearchEngine> active, Map<CheckBoxPreference, SearchEngine> inactive) {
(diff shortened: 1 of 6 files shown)


aldenml:
Isn't it possible to solve this at the style level? That way would be more easy to use the idea in other parts (and avoid `PreferenceSlaveLinearLayout`)

grzesiekrzaca:
We can't get the status of the preference in a layout itself (or at least I don't know how to do that) and we cannot assign a style to a specific preference (generic) [even if we could, first problem stands]. I think we would need a custom preference class that is able to change its background based on its checked state. Is that solution preferable?

aldenml:
I might be wrong, but I believe this preferences are some sort of list with an adapter. If that's the case, well crafted selector should do the trick, and that could be reused in "My Files" I guess. Did you researched that route?

grzesiekrzaca:
To hack the adapter for that list hijack getView and check if it not select all? Hmm I can try that tomorrow

android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
```diff
@@ -17,8 +17,14 @@
 
 package com.frostwire.android.gui.fragments.preference;
 
+import android.support.v4.content.ContextCompat;
```
gubatron: genuine question, is this ok as 'v4', no need for 'v7'? (trying to learn)

android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
```diff
@@ -141,4 +151,33 @@ private void checkAllEngines(boolean checked) {
             UIUtils.showShortMessage(getView(), R.string.search_preferences_one_engine_checked_always);
         }
     }
+
+    @Override
+    protected RecyclerView.Adapter onCreateAdapter(PreferenceScreen preferenceScreen) {
```
gubatron: oh cool, we're moving to `RecyclerView` already

android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
aldenml: I hope no

aldenml:
@grzesiekrzaca Why do need to setup `android:layout="@layout/view_preference_search_engine"` for each preference?

gubatron:
reviewing functionality (for now) on Android 4.4, 5, 6 and 7.

grzesiekrzaca:
@aldenml 
So that we have one more linear layout (normally there is only one that holds the ripple effect, I'm adding another one that allows me to change the color without destroying ripple effect)

gubatron:
Perhaps this needs a rebase:

This is what I get when I try to install on Android 4.4 (API 19)
```
Error:Execution failed for task ':transformClassesWithDexForBasicDebug'.
> com.android.build.api.transform.TransformException: com.android.ide.common.process.ProcessException: java.util.concurrent.ExecutionException: com.android.dex.DexIndexOverflowException: method ID not in [0, 0xffff]: 65536
```

It's hitting the 65k symbol limit.

Let's see on Android 5.

aldenml:
@gubatron for android 4 you need to disable "Instant Run"

gubatron:
that worked! thanks

gubatron:
works fine on Android 4.4

gubatron:
all good on Android 5.1

android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
```diff
@@ -141,4 +151,33 @@ private void checkAllEngines(boolean checked) {
...
+            this.checkedDrawableId = checkedDrawableId;
+            this.unCheckedDrawableId = unCheckedDrawableId;
+        }
+
+        @Override
+        public void onBindViewHolder(PreferenceViewHolder holder, int position) {
+            final CheckBoxPreference preference = (CheckBoxPreference) getItem(position);
+
+            preference.onBindViewHolder(holder);
+            if (!preference.getKey().equals("frostwire.prefs.search.preference_category.select_all")) {
```
gubatron: isn't there a `Constants.PREF_KEY_...` value for this?
There should be one, we don't usually have preferences like this hardcoded in the code.
If you find any, let's make sure we use `Constants.PREF_KEY_...` values

gubatron: I've gone ahead and refactored this for you (sending a Pull Request)

However, I'm not sure why it's necessary to have this as a setting, as the state of this checkbox can be set programmatically on the fly if all the search engines available are checked.

gubatron:
All good on Android 6.0.

Sent you a PR with refactors
https://github.com/grzesiekrzaca/frostwire/pull/2

I'll be waiting for an explanation on why the added complexity of a new setting. I think it's unnecessary.

gubatron:
ok, I think the explanation will be that this was the easiest way to add that checkbox in there, instead of adding some hack to the layout.

If that's the case, LGTM @aldenml

gubatron:
@grzesiekrzaca if @aldenml is good with this, let's not just merge.
Let's rebase all this, then squash into a single commit before merging.

aldenml:
Well
1- I don't think `view_preference_search_engine.xml` is necessary. You should be able to craft a good selector and setting it as background in the adapter.
2- Since the key `"frostwire.prefs.search.preference_category.select_all"` is necessary for the lookup, that should be a constant inside this class `SearchFragment` (not in `Constants`). This is a "key reference", not a preference we can access outside the preference UI.
3- You don't need `android.support.v4.content.ContextCompat`
4- I suppose you can make `CheckedAwarePreferenceGroupAdapter` static.

gubatron:
>2- Since the key "frostwire.prefs.search.preference_category.select_all" is necessary for the lookup, that should be a constant inside this class SearchFragment (not in Constants). This is a "key reference", not a preference we can access outside the preference UI.

good point.

gubatron:
Points 2 (constant) and 4 (static class) [addressed here](https://github.com/grzesiekrzaca/frostwire/pull/2/commits/00832dbe1ab6ea76bfd536f9d7edce406b3e4e9e) - (PR for @grzesiekrzaca's branch)

grzesiekrzaca:
@aldenml 
1- well here's the problem:
a) if I change the background with another ripple and a color the hotspot gets deleted - the ripple is started from the top left corner.
```
<ripple xmlns:android="http://schemas.android.com/apk/res/android" android:color="@color/basic_gray_dark">
    <item android:drawable="@color/selected_search_background" />
</ripple>
```
(and another with different color)
I tried to get a touch event and set hotspot there but that doesn't work. 
b) selector won't work because there is no state_checked to be processed in linear layout (this was the reason I had the custom linear layout before)

android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
```diff
@@ -141,4 +155,32 @@ private void checkAllEngines(boolean checked) {
...
+    }
+
+    private static class CheckedAwarePreferenceGroupAdapter extends PreferenceGroupAdapter {
+        final int checkedDrawableId;
+        final int unCheckedDrawableId;
+        final WeakReference<Activity> activityRef;
+
+        CheckedAwarePreferenceGroupAdapter(Activity activity, PreferenceGroup preferenceGroup, int checkedDrawableId, int unCheckedDrawableId) {
+            super(preferenceGroup);
+            activityRef = Ref.weak(activity);
```
aldenml: Why do we need an activity here?

android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
```diff
@@ -141,4 +155,32 @@ private void checkAllEngines(boolean checked) {
...
+            this.unCheckedDrawableId = unCheckedDrawableId;
+        }
+
+        @Override
+        public void onBindViewHolder(PreferenceViewHolder holder, int position) {
+            final CheckBoxPreference preference = (CheckBoxPreference) getItem(position);
+            preference.onBindViewHolder(holder);
+            if (!preference.getKey().equals(PREF_KEY_SEARCH_SELECT_ALL) && Ref.alive(activityRef)) {
+                int drawableId = preference.isChecked() ? checkedDrawableId : unCheckedDrawableId;
+                holder.itemView.setBackground(activityRef.get().getResources().getDrawable(drawableId));
```
aldenml: can't the resources be set by id?

aldenml:
@grzesiekrzaca ok, I suppose we can take a deep look at it later. I left a couple of comments that I need you to address before merge

android/src/com/frostwire/android/gui/fragments/preference/SearchFragment.java
```diff
@@ -177,9 +172,9 @@ private void checkAllEngines(boolean checked) {
         public void onBindViewHolder(PreferenceViewHolder holder, int position) {
             final CheckBoxPreference preference = (CheckBoxPreference) getItem(position);
             preference.onBindViewHolder(holder);
-            if (!preference.getKey().equals(PREF_KEY_SEARCH_SELECT_ALL) && Ref.alive(activityRef)) {
+            if (!preference.getKey().equals(PREF_KEY_SEARCH_SELECT_ALL)) {
                 int drawableId = preference.isChecked() ? checkedDrawableId : unCheckedDrawableId;
-                holder.itemView.setBackground(activityRef.get().getResources().getDrawable(drawableId));
+                holder.itemView.setBackgroundResource(drawableId);
```
gubatron: oh cool, forgot about this method

PR #336, opened 2017-01-24
[android] fixed styling and padding on API 17-19

marcelinkaaa:
@aldenml I know it is adding a new v-17 themes file - but it is the most straightforward solution I found to control the paddings of the list in preferences. Below are before screenshots from api17, 18,19. 
API 16 was good with only the layout. Let me know if we can keep it. 

screen shot 2017-01-24 at 8 21 07 am
screen shot 2017-01-24 at 8 34 41 am
screen shot 2017-01-24 at 8 32 16 am

### DIFF
--- a/android/res/layout/view_preference_category_material.xml
+++ b/android/res/layout/view_preference_category_material.xml
@@ -22,8 +22,7 @@
     android:layout_height="wrap_content"
     android:layout_marginBottom="16dip"
     android:textAppearance="@style/Preference_TextAppearanceMaterialBody2"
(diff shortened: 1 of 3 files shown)


marcelinkaaa:
Those are after pics. 
screen shot 2017-01-24 at 8 54 39 am
screen shot 2017-01-24 at 9 01 02 am
screen shot 2017-01-24 at 8 59 31 am

aldenml:
Thanks! I was working on it and this seems good

PR #344, opened 2017-02-06
Change of 3g/4g option to wifi only option 

grzesiekrzaca:


aldenml:
Well, I feel very uneasy with the new `interface Migration` :)

aldenml:
Seems reasonable, only that maybe you can figure out a more particular way of handle the migration inside `ConfigurationDefaults` without introducing a new type.

grzesiekrzaca:
I was thinking about future - we might need that(?) and seemed like a place to plug those in might be handy.
But as we have only one now should I just collapse it into a single function? 
```
private ConfigurationManager(Application context) {
          preferences = PreferenceManager.getDefaultSharedPreferences(context);
          editor = preferences.edit();
          defaults = new ConfigurationDefaults();
          initPreferences();
          migrateWifiOnlyPreference();//here we check status and perform this migration
      }
```

aldenml:
I rather put the logic inside the setup for that preference in `ConfigurationDefaults`

android/src/com/frostwire/android/core/Constants.java
```diff
@@ -52,7 +52,13 @@ private Constants() {
     public static final String PREF_KEY_CORE_LAST_SEEN_VERSION = "frostwire.prefs.core.last_seen_version";
 
     public static final String PREF_KEY_NETWORK_ENABLE_DHT = "froswire.prefs.network.enable_dht";
+    /**
+     * @deprecated left for migration purposes, use {@link #PREF_KEY_NETWORK_USE_WIFI_ONLY }
+     */
+    @Deprecated
     public static final String PREF_KEY_NETWORK_USE_MOBILE_DATA = "frostwire.prefs.network.use_mobile_data";
+    public static final String PREF_MIGRATION_FINISHED_KEY_NETWORK_USE_MOBILE_DATA_TO_USE_WIFI_ONLY = "frostwire.prefs.migration.use_mobile_data.use_wifi_only";
```
aldenml: and this looks weird

grzesiekrzaca: Well, we need the migration to run once. So somehow we need to mark that it's been completed use the new preference

android/src/com/frostwire/android/core/ConfigurationManager.java
```diff
@@ -62,13 +65,24 @@ public static ConfigurationManager instance() {
...
+    /**
+     * This method migrates the deprecated {@link Constants#PREF_KEY_NETWORK_USE_MOBILE_DATA}
+     * to the new {@link Constants#PREF_KEY_NETWORK_USE_WIFI_ONLY}
+     * it will be run only once
+     */
+    private void migrateWifiOnlyPreference() {
+        final String MIGRATION_MARKER = "frostwire.prefs.migration.use_mobile_data.use_wifi_only";
+        if(!getBoolean(MIGRATION_MARKER)) {
+            setBoolean(Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY, !getBoolean(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA));
+            setBoolean(MIGRATION_MARKER, true);
```
gubatron: Q: can preferences be deleted?

If so, the marker is unnecessary.

aldenml: I don't think that marker is necessary either

gubatron: To delete a preference:

```
SharedPreferences.Editor.remove(String key)
SharedPreferences.Editor.commit()
```

Then you can get rid of that `MIGRATION_MARKER` and eliminate the added complexity.

android/src/com/frostwire/android/gui/fragments/preference/ApplicationFragment.java
```diff
@@ -67,9 +71,40 @@ public ApplicationFragment() {
...
+                    turnOffTransfers();
+                }
+                return true;
+            }
+        });
+    }
+
+    private void turnOffTransfers() {
+        TransferManager.instance().stopHttpTransfers();
+        TransferManager.instance().pauseTorrents();
```
gubatron: perhaps here we should stop the bittorrent connection altogether as there might still be DHT/Local Discovery related traffic

aldenml: those packets will not travel if not in Wi-Fi, not clear what happens in network reconnection

android/src/com/frostwire/android/core/ConfigurationManager.java
gubatron: Here's a [PR](https://github.com/grzesiekrzaca/frostwire/pull/6), let's get this over with

android/src/com/frostwire/android/gui/fragments/preference/ApplicationFragment.java
gubatron: I think what happens now is that transfers won't start by themselves again.
The ideal thing I guess would be to keep track of which transfers were active and then bring them back, but right now, none are being kept anywhere. That could be the subject of another ticket once this is merged.

My fix to remove the marker logic was merged, but this now needs to be rebased.

grzesiekrzaca:
rebased

aldenml:
Did you add `Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY` to `ConfigurationDefaults#load()`?

grzesiekrzaca:
done

android/src/com/frostwire/android/core/ConfigurationDefaults.java
```diff
@@ -102,6 +102,7 @@ private void load() {
 
         defaultValues.put(Constants.PREF_KEY_NETWORK_ENABLE_DHT, true);
         defaultValues.put(Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA, true);
+        defaultValues.put(Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY, false);
```
aldenml: don't you want to remove `Constants.PREF_KEY_NETWORK_USE_MOBILE_DATA` from here? and maybe move the constant to a more local place?

android/src/com/frostwire/android/gui/fragments/preference/ApplicationFragment.java
```diff
@@ -70,9 +72,39 @@ protected void initComponents() {
...
 
+    private void setupDataSaving() {
+        SwitchPreferenceCompat preference = findPreference(Constants.PREF_KEY_NETWORK_USE_WIFI_ONLY);
+        preference.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
+            @Override
+            public boolean onPreferenceChange(final Preference preference, Object newValue) {
+                boolean newVal = (Boolean) newValue;
+                if (newVal && !NetworkManager.instance().isDataWIFIUp()) {
+                    if (TransferManager.instance().isHttpDownloadInProgress()) {
+                        UIUtils.showYesNoDialog(getActivity(), R.string.data_saving_kill_http_warning, R.string.data_saving_kill_http_warning_title, new DialogInterface.OnClickListener() {
```
aldenml: I don't think this `UIUtils.showYesNoDialog` is safe anymore. Did you test rotation and context leaks? Take a look at `KitKatStoragePreference` for an example of how to use `YesNoDialog`

android/src/com/frostwire/android/gui/activities/SettingsActivity.java
```diff
@@ -127,7 +127,7 @@ private void switchToFragment(String fragmentName, Bundle args, CharSequence tit
         Fragment f = Fragment.instantiate(this, fragmentName, args);
         FragmentTransaction transaction = getFragmentManager().beginTransaction();
         transaction.setTransition(FragmentTransaction.TRANSIT_FRAGMENT_FADE);
-        transaction.replace(R.id.activity_settings_content, f);
+        transaction.replace(R.id.activity_settings_content, f, fragmentName);
```
aldenml: why do you need this?

grzesiekrzaca: so that the fragment has a tag and is found with `getFragments()` when `dispatchDialogClick` from in `AbstractDialog` is called

```
 if (activity instanceof AbstractActivity) {
            List<Fragment> fragments = ((AbstractActivity) activity).getFragments();  //here

            for (Fragment f : fragments) {
                dispatchDialogClickSafe(f, tag, which);
            }
        }
```

aldenml: I see

grzesiekrzaca: Took me a while before figuring this out.

android/res/xml/settings_application.xml
```diff
@@ -25,12 +25,11 @@
...
             android:summary="@string/bittorrent_network_summary"
             android:title="@string/bittorrent" />
-        <!--  Wi-Fi Networks Only should replace Use 3G/4G for BitTorrent -->
-        <android.support.v7.preference.SwitchPreferenceCompat
-            android:key="frostwire.prefs.network.use_mobile_data"
-            android:summary="@string/use_mobile_data_summary"
-            android:title="@string/use_mobile_data" />
         <android.support.v7.preference.SwitchPreferenceCompat
+            android:key="frostwire.prefs.network.use_wifi_only"
+           android:summary="@string/wifi_networks_only_summary"
```
aldenml: formatting here (not important now)

gubatron:
sweet candy crush

PR #348, opened 2017-02-08
[android] added _selected blue icons for action_favorite

marcelinkaaa:
Also one remaining ic_action_favorite icon was still in grey color. Changed to already used white.


aldenml:
thank you

Issue #166, opened 2016-07-12
[desktop/android] New Advanced Setting to not start BitTorrent engine unless VPN is active


gubatron:
Per [suggestion on reddit](https://www.reddit.com/r/Bittorrent/comments/4n4vjw/how_to_prevent_accidental_starting_of_bittorrent/) of [/user/Tyisgom](https://www.reddit.com/user/Tyisgom)

grzesiekrzaca:
Only the android part of the issue is done. Desktop is still in progress (see my slack message)

PR #349, opened 2017-02-09
Feature/bittorrent only on vpn desktop

grzesiekrzaca:
The second part of #166.
Not sure if this is a correct way.

### DIFF
--- a/desktop/lib/messagebundles/default.po
+++ b/desktop/lib/messagebundles/default.po
@@ -4273,8 +4273,19 @@ msgstr ""
 msgid "Require VPN connection for BitTorrent"
 msgstr ""
 
-#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/options/panes/TorrentConnectionPaneItem.java:42
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/frostwire/gui/bittorrent/TorrentFetcherDownload.java:249
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/frostwire/gui/bittorrent/TorrentFetcherDownload.java:273
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/frostwire/gui/bittorrent/BTDownloadMediator.java:611
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/frostwire/gui/bittorrent/BTDownloadMediator.java:676
 msgid "VPN is inactive. Current settings require active VPN connection before starting BitTorrent engine. Setup a VPN connection or check advanced settings."
+msgstr ""
+
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/StatusLine.java:391
+msgid "BitTorrent is disabled!"
+msgstr ""
+
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/StatusLine.java:392
+msgid "Due to current settings without VPN connection BitTorrent will not start. Click to see the settings screen."
 msgstr ""
 
 #: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/options/panes/TorrentGlobalSpeedPaneItem.java:34
--- a/desktop/lib/messagebundles/default.po
+++ b/desktop/lib/messagebundles/default.po
@@ -4269,6 +4269,14 @@ msgstr ""
 msgid "Enable Distributed Hash Table (DHT)"
 msgstr ""
 
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/options/panes/TorrentConnectionPaneItem.java:42
+msgid "Require VPN connection for BitTorrent"
+msgstr ""
+
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/options/panes/TorrentConnectionPaneItem.java:42
+msgid "VPN is inactive. Current settings require active VPN connection before starting BitTorrent engine. Setup a VPN connection or check advanced settings."
+msgstr ""
+
 #: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/options/panes/TorrentGlobalSpeedPaneItem.java:34
 msgid "BitTorrent Global Tranfer Speeds"
 msgstr ""
--- a/desktop/src/com/frostwire/gui/bittorrent/BTDownloadMediator.java
+++ b/desktop/src/com/frostwire/gui/bittorrent/BTDownloadMediator.java
@@ -38,20 +38,14 @@ import com.frostwire.search.youtube.YouTubeCrawledSearchResult;
 import com.frostwire.transfers.TransferState;
 import com.frostwire.util.HttpClientFactory;
 import com.frostwire.util.http.HttpClient;
-import com.limegroup.gnutella.gui.GUIMediator;
-import com.limegroup.gnutella.gui.I18n;
-import com.limegroup.gnutella.gui.PaddedPanel;
+import com.limegroup.gnutella.gui.*;
 import com.limegroup.gnutella.gui.actions.LimeAction;
-import com.limegroup.gnutella.gui.iTunesMediator;
 import com.limegroup.gnutella.gui.search.GenericCellEditor;
 import com.limegroup.gnutella.gui.tables.AbstractTableMediator;
 import com.limegroup.gnutella.gui.tables.LimeJTable;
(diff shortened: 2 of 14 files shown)


desktop/lib/messagebundles/default.po
```diff
@@ -4269,6 +4269,25 @@ msgstr ""
 msgid "Enable Distributed Hash Table (DHT)"
 msgstr ""
 
+#: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/options/panes/TorrentConnectionPaneItem.java:42
```
aldenml: I believe these `*.po` are generated by angel at the end for translation

aldenml:
I didn't tested, but it looks reasonable, only because my main reservations are in the core `VPNs` class.

desktop/lib/messagebundles/default.po
grzesiekrzaca: Yeah, I was afraid of that. This direct path was suspicious. I asked him on slack about that, still waiting on a response.

gubatron: the translation files are all generated from the `I18n.tr("some string")` code calls all over the project. Using `gradle gettextXXXX` tasks.

```
gettextExtract -> extracts all the I18n.tr() strings, into the default file
gettextBundle -> grabs all the other .po files and creates a messages.jar file
```

We only touch the other language files, not the default (english one)

desktop/src/com/frostwire/gui/bittorrent/BTDownloadMediator.java
```diff
@@ -674,7 +672,11 @@ public void run() {
                     }
 
                     if (saveDir == null) {
-                        BTEngine.getInstance().download(torrentFile, null, filesSelection);
+                        if (ConnectionSettings.MANDATORY_VPN_FOR_BITTORRENT.getValue() && !VPNs.isVPNActive()) {
```
gubatron: I guess this check should be moved before this `if (saveDir == null)` altogether (and add a `return`), otherwise the `openTorrentForSeed` case will occur.

I guess I'd move it all the way to the start of the method as a short-circuit mechanism.

desktop/src/com/frostwire/gui/bittorrent/TorrentFetcherDownload.java
```diff
@@ -242,7 +245,11 @@ private void downloadTorrent(final byte[] data, final List<TcpEndpoint> peers) {
             try {
                 TorrentInfo ti = TorrentInfo.bdecode(data);
                 boolean[] selection = calculateSelection(ti, relativePath);
-                BTEngine.getInstance().download(ti, null, selection, peers);
+                if (ConnectionSettings.MANDATORY_VPN_FOR_BITTORRENT.getValue() && !VPNs.isVPNActive()) {
+                    GUIMediator.showWarning(I18n.tr("VPN is inactive. Current settings require active VPN connection before starting BitTorrent engine. Setup a VPN connection or check advanced settings."));
+                } else {
+                    BTEngine.getInstance().download(ti, null, selection, peers);
+                }
```
gubatron: so I see a pattern here with this `if/else`, let's refactor this into a method that takes all the parameters as the `BTEngine.getInstance().download()` method takes on the else. #DRY

desktop/src/com/frostwire/gui/updates/UpdateManager.java
```diff
@@ -210,16 +212,16 @@ private void handlePossibleUpdateMessage(UpdateMessageReader umr, boolean force)
 
             // Logic for Windows or Mac Update
             if (OSUtils.isWindows() || OSUtils.isMacOSX()) {
-                if (hasUrl && !hasTorrent && !hasInstallerUrl) {
+                if ((hasUrl && !hasTorrent && !hasInstallerUrl) || hasUrl && !canUseTorrent()) {
```
gubatron: let's please remove this check for the update manager. This one torrent is a system torrent, should be fine.

desktop/src/com/limegroup/gnutella/gui/StatusLine.java
```diff
@@ -457,7 +480,7 @@ private void updateCenterPanel() {
 
     private void updateSeedingStatus() {
         boolean seedingStatus = SharingSettings.SEED_FINISHED_TORRENTS.getValue();
-        seedingStatusButton.setText("<html>"+(seedingStatus ? I18n.tr("<b>Seeding</b>") : I18n.tr("<b>Not Seeding</b>"))+"</html>");
+        seedingStatusButton.setText("<html>" + (seedingStatus ? I18n.tr("<b>Seeding</b>") : I18n.tr("<b>Not Seeding</b>")) + "</html>");
```
gubatron: let's use this opportunity to move the `<b>` tag out of the `I18n.tr` call, just leave the `Seeding` in.

desktop/src/com/limegroup/gnutella/gui/StatusLine.java
```diff
@@ -71,6 +72,11 @@
     private JLabel _bandwidthUsageDown;
     private JLabel _bandwidthUsageUp;
 
+    /**
+     * Shows when bittorrent is disabled due to VPN settings
+     */
+    private VPNBitTorrentDisabledWarningLabel _bittorrent_vpn;
```
gubatron: #style: you've been coding in python lately? ;)
Let's rename to `bittorrentVpn`

luckily in java we have `private`,`public`, `protected` and we don't need to resort to `__mFoo`, `_mBar` naming.

desktop/src/com/limegroup/gnutella/gui/StatusLine.java
```diff
@@ -71,6 +72,11 @@
...
+
     private IconButton _twitterButton;
```
gubatron: and this is here because I was probably coding in python back then :)

desktop/src/com/limegroup/gnutella/gui/VPNStatusRefresher.java
```diff
@@ -0,0 +1,79 @@
...
+            Thread vpnStatusCheckerThread = new Thread("VPNStatus-checker") {
+                @Override
+                public void run() {
+                    //possibly blocking code
+                    final boolean isVPNActive = VPNs.isVPNActive();
+                    GUIMediator.safeInvokeLater(new Runnable() {
+                        @Override
+                        public void run() {
+                            for (VPNStatusListener client : clients) {
+                                client.onStatusUpdated(isVPNActive);
```
gubatron: let's wrap this with a try/catch throwable on this line, It's never good to trust listener implementations.

desktop/src/com/limegroup/gnutella/gui/VPNStatusRefresher.java
```diff
@@ -0,0 +1,79 @@
...
+                            }
+                        }
+                    });
+                }
+            };
+            pool.execute(vpnStatusCheckerThread);
+        }
+    }
+
+    interface VPNStatusListener {
```
gubatron: 👍 

Also thought something like this would be necessary for this feature.

gubatron:
I'll test now, perhaps the notification component I'm thinking is that StatusLine you found @grzesiekrzaca

desktop/lib/messagebundles/default.po
marcelinkaaa: @gubatron If you are talking about the yellow one - yes the layout is called view_rich_notification.xml

gubatron: @marcelinkaaa nope, we're talking about desktop here. I think there was some StatusLine component already there he found, I've not yet been able to test this branch.

desktop/src/com/frostwire/gui/updates/UpdateManager.java
grzesiekrzaca: Well, we would not be entirely true with the option then. Maybe the user doesn't want any BT traffic. Even "safe" stuff. Also making the engine work just for that might be tricky. Would require a lot of reading of the code. (I don't know if running the `SessionManager` leaves any artifacts (JNI part))

aldenml:
Between `VPNBitTorrentGuard` and the "refresher" I'm starting to feel this a little odd, but still inside the limits

grzesiekrzaca:
Yeah I'm not entirely happy with that but had no idea where to put it.

gubatron:
testing, not fully functional. (e.g. when Bittorrent is disabled, prior seeds, do not pause, they stay seeding)

screen shot 2017-02-10 at 9 20 23 am

I'll try to fix a few things, for instance, the label that says that Bittorrent is disabled lags with respect to the VPN status button, ideally these two should change at the same time, I'll see why this is.

gubatron:
Fixing the issue where the components don't refresh at the same time.
It's architectural.

I'm removing the registration and call to VPNStatusRefresher to refresh from inside these components, since this is a blocking call, one is updated, then the other one calls refresh again, then it gets updated.

I'm going to make a single call from outside, and they just listen to the status change.

PR #350, opened 2017-02-09
[android] fix crash due to old class ref

grzesiekrzaca:
this fixes crash introduced in 
https://github.com/frostwire/frostwire/commit/4ceac17a8e6bc5d21f0be5be3420bd2007b128bb
due to renaming of `ButtonActionPreference2` to `ButtonActionPreference`


aldenml:
Thanks

Issue #357, opened 2017-02-24
[android/music player] Equalizer option menu bug 


gubatron:
Jan 24 2016

Tested on: **Samsung S3 Android 4.3, SGH-T399 4.2.2**
FrostWire for Android **version 1.7.0 (build 232) Basic & Plus**
Screen: **Player menu** 
Test: **Test player menu equalizer"**
**Ref. P-M4**

**Observations:**

The equalizer is not being displayed when playing a song.

--
March 02 2016

Updating issue for FrostWire 1.7.1 (build 237) now tested on Lollipop devices too:

The Equalizer of the device (no a FW integrated equalizer) is displayed only on devices that has Audio Settings option...

So, the issue here is the 'Equalizer' option of FrostWire should appear inactive if the device does not offer that setting, otherwise it seems FW offers an Equalizer that does not work.
--

See `AudioPlayerActivity.java`
https://github.com/frostwire/frostwire/blob/master/android/apollo/src/com/andrew/apollo/ui/activities/AudioPlayerActivity.java#L321

https://github.com/frostwire/frostwire/blob/master/android/apollo/src/com/andrew/apollo/ui/activities/AudioPlayerActivity.java#L393
--
Aug 3 2016
This issue persists on FrostWire 1.7.9 (build 293)

An 'Equalizer' option is displayed only on Android 4.2.2 device and on Android 6 and 5 the option is not displayed even if I install an app that offers 'Equalizer'.
--
Feb 23 2017

This issue persists on FrostWire 1.8.9 - build 242.

On Android 4.2 and 4.3 go to:

My Music > Album / Artists / Recent > Select a song > display the list of songs > click at the top right menu > Equalizer
This option is not opening the Equalizer on the device.

On Android 5 and 6:

Equalizer option is not available although I installed an Equalizer app on the device.

CC: @votaguz

votaguz:
https://github.com/frostwire/frostwire/pull/358

aldenml:
To close when tested in QA

gubatron:
@muckachina please test this one

muckachina:
Sorry I'm late testing this one, I saw it last Monday... is Fixed!! :D 

Tested on **Android 4.2** with an equalizer by default and the option appears on the menu as it was before. The equalizer opens with no issues.

On **Android 5 and 6** I installed an external equalizer-app and the option appeared on FW file-menu and the equalizer opens working properly with FW player. 

Thank you thank you!

PR #358, opened 2017-02-24
[android] Solving equalizer doesn't appear in player activity

votaguz:
Issue #357 : 
Just changed the call from ```startActivity``` to ```startActivityForResult``` method. 

The only guess that i have about this issue, it seems to be related to the integer value passed as expected value internally to ```startActivityForResult```.

I checked android ```Activity``` class, found ```startActivity``` calls ```startActivityForResult``` passing a ```-1``` as expected result but if that value is changed by ```0``` (I think we can use any integer) equalizer is showed with no problems

````
    /**
     * Launch a new activity.  You will not receive any information about when
     * the activity exits.  This implementation overrides the base version,
     * providing information about
     * the activity performing the launch.  Because of this additional
     * information, the {@link Intent#FLAG_ACTIVITY_NEW_TASK} launch flag is not
     * required; if not specified, the new activity will be added to the
     * task of the caller.
     *
     * <p>This method throws {@link android.content.ActivityNotFoundException}
     * if there was no Activity found to run the given Intent.
     *
     * @param intent The intent to start.
     * @param options Additional options for how the Activity should be started.
     * See {@link android.content.Context#startActivity(Intent, Bundle)
     * Context.startActivity(Intent, Bundle)} for more details.
     *
     * @throws android.content.ActivityNotFoundException
     *
     * @see #startActivity(Intent)
     * @see #startActivityForResult
     */
    @Override
    public void startActivity(Intent intent, @Nullable Bundle options) {
        if (options != null) {
            startActivityForResult(intent, -1, options);
        } else {
            // Note we want to go through this call for compatibility with
            // applications that may have overridden the method.
            startActivityForResult(intent, -1);
        }
    }
````

58afc8be4a2fa449978163

### DIFF
--- a/android/apollo/src/com/andrew/apollo/utils/NavUtils.java
+++ b/android/apollo/src/com/andrew/apollo/utils/NavUtils.java
@@ -95,7 +95,8 @@ public final class NavUtils {
         try {
             final Intent effects = new Intent(AudioEffect.ACTION_DISPLAY_AUDIO_EFFECT_CONTROL_PANEL);
             effects.putExtra(AudioEffect.EXTRA_AUDIO_SESSION, MusicUtils.getAudioSessionId());
-            context.startActivity(effects);
+            // No result expected for this activity
+            context.startActivityForResult(effects, 0);
         } catch (final ActivityNotFoundException notFound) {
             AppMsg.makeText(context, context.getString(R.string.no_effects_for_you),
                     AppMsg.STYLE_ALERT);


aldenml:
This is good, it's documented in https://developer.android.com/reference/android/media/audiofx/AudioEffect.html#ACTION_DISPLAY_AUDIO_EFFECT_CONTROL_PANEL

> The calling application must use the **startActivityForResult(Intent, int)** method to launch the control panel so that its package name is indicated and used by the control panel application to keep track of changes for this particular application.

gubatron:
shaquille-o-neal

way to go @votaguz !!!

I think @muckachina will be happy to see this ancient issue finally resolved.

muckachina:
Thank you! Issue closed on FW public repo and QA :)

Tested with internal equalizer on Android 4.2 and external equalizer apps on Android 5 and 6... although no all Equalizers are recognized by FW but it seems because they also act like Players (for example 'Music Equalizer' which is a popular one)

gubatron:
kvh2fyopyztshq_rpwiuwnwneuvavxd-ccjr46o0ebe
"Yeah!"

PR #364, opened 2017-02-27
Minor styling adjustments for toolbar and nav

marcelinkaaa:
After adding new toolbar, the drop down options changed its styling from light background to dark background, which we do not use throughout the app. 

screenshot_20170227-131833


android/res/layout/toolbar_main.xml
```diff
@@ -26,6 +26,7 @@
     android:layout_height="48dp"
     android:background="@color/action_bar_background"
     android:theme="@style/ThemeOverlay.AppCompat.Dark.ActionBar"
+    app:popupTheme="@style/ThemeOverlay.AppCompat.Light"
```
aldenml: interesting

PR #365, opened 2017-02-27
[android] added linearlayout anchor for the bittorrent snackbar

marcelinkaaa:
snackbar is now appearing on top of the android bottom nav, works


aldenml:
you should use more `findView`

PR #373, opened 2017-03-03
[desktop] Added piece size selector for create torrent window

votaguz:
Added piece size selector for create torrent window.

### DIFF
--- a/desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
+++ b/desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
@@ -37,10 +37,7 @@ import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 import javax.swing.filechooser.FileFilter;
 import java.awt.*;
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
-import java.awt.event.MouseAdapter;
-import java.awt.event.MouseEvent;
+import java.awt.event.*;
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
@@ -98,6 +95,24 @@ public class CreateTorrentDialog extends JDialog {
     private JScrollPane _textTrackersScrollPane;
     private String _invalidTrackerURL;
     private JButton _buttonClose;
+
+    private JComboBox _pieceSizeComboBox;
+    private JLabel _labelPieceSize;
+    private int pieceSize;
+    private static final String[] PIECE_SIZE_OPTIONS = {
+            "(Auto Detect)",
+            "16 kB",
+            "32 kB",
+            "64 kB",
+            "128 kB",
+            "256 kB",
+            "512 kB",
+            "1024 kB",
+            "2048 kB",
+            "4096 kB"
+    };
+
+    private static final int[] PIECE_SIZES = {0, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096};
 
     public CreateTorrentDialog(JFrame frame) {
         super(frame);
@@ -182,7 +197,7 @@ public class CreateTorrentDialog extends JDialog {
 
     private void initTorrentTracking() {
         JPanel torrentTrackingPanel = new JPanel(new MigLayout("fill"));
-        GUIUtils.setTitledBorderOnPanel(torrentTrackingPanel, I18n.tr("Tracking"));
+        GUIUtils.setTitledBorderOnPanel(torrentTrackingPanel, I18n.tr("Torrent Properties"));
 
         _checkUseDHT = new JCheckBox(I18n.tr("Trackerless Torrent (DHT)"), true);
         _checkUseDHT.setToolTipText(I18n.tr("Select this option to create torrents that don't need trackers, completely descentralized. (Recommended)"));
@@ -212,6 +227,12 @@ public class CreateTorrentDialog extends JDialog {
         _textWebSeeds = new JTextArea(4, 70);
         ThemeMediator.fixKeyStrokes(_textWebSeeds);
         torrentTrackingPanel.add(new JScrollPane(_textWebSeeds), "gapright 5, gapleft 80, gapbottom 5, hmin 165px, growx 60, growy");
+
+        // Select piece size
+        _labelPieceSize = new JLabel(I18n.tr("Piece Size"));
+        torrentTrackingPanel.add(_labelPieceSize, "align left, pushy, gapleft 5");
+        _pieceSizeComboBox = new JComboBox(PIECE_SIZE_OPTIONS);
+        torrentTrackingPanel.add(_pieceSizeComboBox, "align left, pushx, gapright 80");
 
         //suggest DHT by default 
         updateTrackerRelatedControlsAvailability(true);
@@ -286,6 +307,16 @@ public class CreateTorrentDialog extends JDialog {
                 }
             }
         });
+
+        _pieceSizeComboBox.addActionListener(
+                new ActionListener() {
+                    @Override
+                    public void actionPerformed(ActionEvent e) {
+                        pieceSize = PIECE_SIZES[_pieceSizeComboBox.getSelectedIndex()];
+                        System.out.println(pieceSize);
+                    }
+                }
+        );
     }
 
     private void updateTrackerRelatedControlsAvailability(boolean useDHT) {
@@ -570,7 +601,7 @@ public class CreateTorrentDialog extends JDialog {
             reportCurrentTask(I18n.tr("Adding files..."));
             libtorrent.add_files(fs, f.getPath());
 
-            create_torrent torrent = new create_torrent(fs);
+            create_torrent torrent = new create_torrent(fs, pieceSize);
             torrent.set_priv(false);
             torrent.set_creator("FrostWire " + FrostWireUtils.getFrostWireVersion() + " build " + FrostWireUtils.getBuildNumber());
 
--- a/desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
+++ b/desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
@@ -126,7 +126,7 @@ public class CreateTorrentDialog extends JDialog {
             return humanRep;
         }
     }
-
+    private static final PieceSize[] values = PieceSize.values();
 
     public CreateTorrentDialog(JFrame frame) {
         super(frame);
@@ -245,7 +245,7 @@ public class CreateTorrentDialog extends JDialog {
         JLabel labelPieceSize = new JLabel(I18n.tr("Piece Size"));
         torrentTrackingPanel.add(labelPieceSize, "aligny top, pushy, gapleft 5, gapright 10, wmin 150px");
 
-        pieceSizeComboBox = new JComboBox(PieceSize.values());
+        pieceSizeComboBox = new JComboBox(values);
         torrentTrackingPanel.add(pieceSizeComboBox, "gapright 5, gapleft 80, width 175px");
 
         //suggest DHT by default 
@@ -333,7 +333,9 @@ public class CreateTorrentDialog extends JDialog {
     }
 
     private void onPieceSizeSelected(int selectedIndex) {
-        pieceSize = PieceSize.values()[selectedIndex].kb*_1024BYTES;
+        if (selectedIndex > 0 && values.length > selectedIndex && values[selectedIndex] != null) {
+            pieceSize = values[selectedIndex].kb*_1024BYTES;
+        }
     }
 
     private void updateTrackerRelatedControlsAvailability(boolean useDHT) {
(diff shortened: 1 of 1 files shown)


desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -99,6 +96,24 @@
     private String _invalidTrackerURL;
     private JButton _buttonClose;
 
+    private JComboBox _pieceSizeComboBox;
+    private JLabel _labelPieceSize;
+    private int pieceSize;
+    private static final String[] PIECE_SIZE_OPTIONS = {
```
votaguz: I'm not sure if this is the proper way to do it. Since JComboBox constructor accepts just unidimensional iterables  i have to do it like this.  I was wondering if i could do it using a data structure Key, Value type, but after think it a little bit found that i will need a Array/Vector to populate selector options.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -99,6 +96,24 @@
...
+            "64 kB",
+            "128 kB",
+            "256 kB",
+            "512 kB",
+            "1024 kB",
+            "2048 kB",
+            "4096 kB"
+    };
+
+    private static final int[] PIECE_SIZES = {0, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096};
```
votaguz: Same as before

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -570,7 +601,7 @@ boolean makeTorrent() {
             reportCurrentTask(I18n.tr("Adding files..."));
             libtorrent.add_files(fs, f.getPath());
 
-            create_torrent torrent = new create_torrent(fs);
+            create_torrent torrent = new create_torrent(fs, pieceSize);
```
votaguz: I did this because found that jlibtorrent has a constructor for this situation.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -213,6 +228,12 @@ private void initTorrentTracking() {
         ThemeMediator.fixKeyStrokes(_textWebSeeds);
         torrentTrackingPanel.add(new JScrollPane(_textWebSeeds), "gapright 5, gapleft 80, gapbottom 5, hmin 165px, growx 60, growy");
 
+        // Select piece size
+        _labelPieceSize = new JLabel(I18n.tr("Piece Size"));
+        torrentTrackingPanel.add(_labelPieceSize, "align left, pushy, gapleft 5");
```
votaguz: Elements aren't positioned yet, because i still don't discover how to use coordinates system in MigLayout.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
gubatron: You can pass a `ComboBoxModel<E>`, which can return non-unidimensional objects.
https://docs.oracle.com/javase/7/docs/api/javax/swing/ComboBoxModel.html

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
gubatron: you discovered this "by accident"

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
gubatron: google MigLayout CheatSheet

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -99,6 +96,24 @@
     private String _invalidTrackerURL;
     private JButton _buttonClose;
 
+    private JComboBox _pieceSizeComboBox;
+    private JLabel _labelPieceSize;
+    private int pieceSize;
+    private static final String[] PIECE_SIZE_OPTIONS = {
+            "(Auto Detect)",
```
gubatron: Please make this string translatable:
```java
"(" + I18n.tr("Auto Detect") + ")",
```

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -99,6 +96,24 @@
     private String _invalidTrackerURL;
     private JButton _buttonClose;
 
+    private JComboBox _pieceSizeComboBox;
+    private JLabel _labelPieceSize;
```
gubatron: let's not use pythonic style, remove the prefix "_"

gubatron: on both variables.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -286,6 +307,16 @@ public void mouseClicked(MouseEvent e) {
                 }
             }
         });
+
+        _pieceSizeComboBox.addActionListener(
+                new ActionListener() {
+                    @Override
+                    public void actionPerformed(ActionEvent e) {
+                        pieceSize = PIECE_SIZES[_pieceSizeComboBox.getSelectedIndex()];
```
gubatron: move this code outside into a function (of `CreateTorrentDialog`) called
```java
private void onPieceSizeSelected(int selectedIndex)
```

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -99,6 +96,24 @@
     private String _invalidTrackerURL;
     private JButton _buttonClose;
 
+    private JComboBox _pieceSizeComboBox;
```
gubatron: no need to keep a reference for this, make this variable locally declared

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
gubatron: this label can also be locally declared, no need to keep as a member.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
gubatron: Q: Does jlibtorrent receive this parameter in kB? or in bytes? where you able to test with a torrent and see how how many pieces you would get. Say, make a dummy file of exactly 32kb (32*1024) and make a torrent with 16kb pieces, do you see 2 piece hashes inside the created .torrent file?

gubatron: I think a cooler way of doing this would be to create an `enum`, instead of 2 arrays

```java
private enum PieceSize {
  AUTO_DETECT(0,"(" + I18ntr("Auto Detect") + ")"),
  _16KB(16), // I include the '_' because enums can't start with numbers.
  _32KB(32),
  ...
  _4096KB;

  private int kb;
  private String humanRep;

  PieceSize(int k, String rep) {
      kb = k;
      humanRep = rep;
  }

  PieceSize(int k) {
      this(k, k + " kB");
  } 
}```

Then create the combobox model using the enums.

Comment (votaguz) [inline desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java]: I did the test creating a dummy file and then creating a torrent from that file, piece sizes should by multiplied by 1024.

File: desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java (hunk 8 above)
Comment (votaguz) [inline desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java]: I leave it because i need that reference to bind the action listener later.

File: desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java [hunk 9]
```diff
@@ -99,6 +94,40 @@
     private String _invalidTrackerURL;
     private JButton _buttonClose;
 
+    private JComboBox pieceSizeComboBox;
+    private int pieceSize;
+    private static final int _1024BYTES = 1024;
+    private enum PieceSize {
```
gubatron: love it

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -99,6 +94,40 @@
     private String _invalidTrackerURL;
     private JButton _buttonClose;
 
+    private JComboBox pieceSizeComboBox;
```
gubatron: did you try making this a local variable?

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -99,6 +94,40 @@
...
+    private int pieceSize;
+    private static final int _1024BYTES = 1024;
```
gubatron: since this is used only once, perhaps we can do without this constant

votaguz: OK, i just wanna make visible in case someone try to figure out why that value.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
votaguz: Yes i tried, but ActionListener don't see the reference.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -211,7 +240,13 @@ private void initTorrentTracking() {
 
         _textWebSeeds = new JTextArea(4, 70);
         ThemeMediator.fixKeyStrokes(_textWebSeeds);
-        torrentTrackingPanel.add(new JScrollPane(_textWebSeeds), "gapright 5, gapleft 80, gapbottom 5, hmin 165px, growx 60, growy");
+        torrentTrackingPanel.add(new JScrollPane(_textWebSeeds), "gapright 5, gapleft 80, gapbottom 5, hmin 165px, growx 60, growy, wrap");
+
+        JLabel labelPieceSize = new JLabel(I18n.tr("Piece Size"));
+        torrentTrackingPanel.add(labelPieceSize, "aligny top, pushy, gapleft 5, gapright 10, wmin 150px");
+
+        pieceSizeComboBox = new JComboBox(values);
```
gubatron: If you declare this here as `final` the actionListener will see the reference.

gubatron: `final JComboBox pieceSizeComboBox = ...`

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -286,6 +321,21 @@ public void mouseClicked(MouseEvent e) {
...
+                    @Override
+                    public void actionPerformed(ActionEvent e) {
+                        onPieceSizeSelected(pieceSizeComboBox.getSelectedIndex());
+                    }
+                }
+        );
+    }
+
+    private void onPieceSizeSelected(int selectedIndex) {
+        if (selectedIndex > 0 && values.length > selectedIndex && values[selectedIndex] != null) {
```
gubatron: there's a bug here.
let me know if you want me to tell you what it is.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
gubatron: I think it's self explanatory because the enum clearly says you're working with`KB`.
If you still want to keep it, you could name it `KB` instead.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
```diff
@@ -286,6 +321,21 @@ public void mouseClicked(MouseEvent e) {
...
+            pieceSize = values[selectedIndex].kb*_1024BYTES;
```
gubatron: another way of doing this, would be to encapsulate the logic of how many bytes this is supposed to be, and not put the responsability on others to do this multiplication if you added a `.bytes()` method in the enum, which would then do the multiplication internally. I think that'd be cleaner.

Then here you'd have

```java
pieceSize = values[selectedIndex].bytes()
```

gubatron: and you could make the `.kb` property `private` inside the enum.
In the end, the user only needs to see the human representation for the combobox.

desktop/src/com/frostwire/gui/bittorrent/CreateTorrentDialog.java
votaguz: Well ```selectedIndex``` can be 0, in that case isn't posible to reach the first value.

aldenml:
Other than @gubatron's comments, it LGTM (even if I'm not a big fan of the `enum`). At some point in the future, this should be refactored to use `TorrentBuilder`.

gubatron:
I think I suggested the enum to teach @votaguz about enums, and he ended up teaching me that enums have a `.values()` method.

I'm merging and fixing the anal quirks I have in my mind.

aldenml:
👍

PR #376, opened 2017-03-03
Old icon cleanup

marcelinkaaa:
There has been quite a bit. Deleted old wi-fi sharing icons, checkbox icons that are now replaced with a widget, old custom number picker icons, etc. Before deleting I check if the icon is still used in the project and run the app to make sure it builds. 
Deleted corresponding no longer used xml selectors that used the old icons.


aldenml:
Nice!

gubatron:
dora-cleanup

PR #380, opened 2017-03-03
Drawable cleanup part 2

marcelinkaaa:
Went through the drawables, and it looked that we are no longer using some of them. Some backgrounds, selectors, etc. from our old action bar, ones that we never used in Apollo (when we customized it a bit to look more like the rest of the app), etc. There was no usages of these throughout the app. After removal the app was built and installed properly.


aldenml:
To confirm, are you doing a full rebuild? (Build->Clean Project)

marcelinkaaa:
ha good point, I was not. I would uninstall the app via terminal and install again. But just did clean project and installed the app from scratch and all is fine.

PR #382, opened 2017-03-06
Feature/google cast support

grzesiekrzaca:
Barebones needed to add Cast support to the project and simple Http server that will serve the content.

### DIFF
--- a/android/src/com/frostwire/android/core/server/CastFileServer.java
+++ b/android/src/com/frostwire/android/core/server/CastFileServer.java
@@ -1,0 +1,216 @@
+package com.frostwire.android.core.server;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import com.frostwire.util.Logger;
+
+import fi.iki.elonen.NanoHTTPD;
+
+public class CastFileServer extends NanoHTTPD {
+
+    private static final int PORT = 8080;
+    private static final int TIMEOUT = 5000;
+
+    private static final String ACCEPT_RANGES_HEADER = "Accept-Ranges";
+    private static final String CONTENT_LENGTH_HEADER = "Content-Length";
+    private static final String CONTENT_RANGE_HEADER = "Content-Range";
+    private static final String E_TAG_HEADER = "ETag";
+
+    public static final Logger LOG = Logger.getLogger(CastFileServer.class);
+
+    private HashMap<String, String> fileLookup = new HashMap<>();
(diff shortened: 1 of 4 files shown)


android/src/com/frostwire/android/core/server/CastFileServer.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * @param path path to remove from streaming
+     */
+    public void removeFileFromCast(String path) {
+        fileLookup.remove(reverseFileLookup.remove(path));
+    }
+
+    private String getIp() {
```
gubatron: I bet we can get this from the BTEngine

android/src/com/frostwire/android/core/server/CastFileServer.java
```diff
@@ -0,0 +1,216 @@
...
+                    response = newFixedLengthResponse(Response.Status.NOT_MODIFIED, mime, "");
+                    response.addHeader(ACCEPT_RANGES_HEADER, "bytes");
+                } else {
+                    response = newChunkedResponse(Response.Status.OK, mime,
+                            new FileInputStream(file));
+                    response.addHeader(ACCEPT_RANGES_HEADER, "bytes");
+                    response.addHeader(CONTENT_LENGTH_HEADER, "" + fileLen);
+                    response.addHeader(E_TAG_HEADER, eTag);
+                }
+            }
```
gubatron: I believe a call to 

```java
try {
    response.close()
} catch (Throwable t) { //close quietly }
``` 

is missing here, otherwise the passed `FileInputStream` remains open... unless whoever calls `serveFile` closes it, but I don't see that in the file anywhere.

android/src/com/frostwire/android/core/server/CastFileServer.java
```diff
@@ -0,0 +1,216 @@
...
+    public void start() {
+        try {
+            this.start(TIMEOUT);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    @Override
+    public Response serve(IHTTPSession session) {
```
gubatron: wondering who does the `.close()` on this `Response` object.

gubatron:
Q: how big is the .apk getting after including the casting framework and the nano httpd server?

android/src/com/frostwire/android/core/server/CastFileServer.java
grzesiekrzaca: I didn't see such method, could you point me to it?

android/src/com/frostwire/android/core/server/CastFileServer.java
grzesiekrzaca: It's inside the nanohttpd. the stream is closed after sending the response.

grzesiekrzaca:
10 529 KB -> 11 647 KB
rougly 1MB (10%) increase

PR #391, opened 2017-03-09
[android] fix for ad-view overlaying on top of the menu

marcelinkaaa:
Currently the navigation drawer ad is overlapping the menu, which causes the last menu item - "Exit" to be hidden under the ad itself. 
There is a few hacky ways to do it with layouts/paddings but they are a bit too much. This was the simplest way to accomplish it. 
I am not sure how to do it any other way. If you have any pointers I will be happy to give them a go. 

screen shot 2017-03-09 at 2 36 01 pm
screen shot 2017-03-09 at 2 36 54 pm

### DIFF
--- a/android/res/layout/view_ad_menuitem.xml
+++ b/android/res/layout/view_ad_menuitem.xml
@@ -29,7 +29,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_centerVertical="true"
-        android:layout_marginTop="22dp"
+        android:layout_marginTop="20dp"
         android:layout_toLeftOf="@+id/view_ad_menu_item_thumbnail"
         android:paddingLeft="12dp">
 
--- a/android/res/menu/nav_view_menu.xml
+++ b/android/res/menu/nav_view_menu.xml
@@ -48,5 +48,13 @@
             android:id="@+id/menu_main_shutdown"
             android:icon="@drawable/menu_icon_exit"
             android:title="Exit"/>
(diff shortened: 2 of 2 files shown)


aldenml:
With the latest two commits in `master` I'm kind of loss at what exactly look for. Is this PR a real fix, or only a workaround? That last item in `nav_view_menu.xml` seems suspicious.

marcelinkaaa:
The other commits concerned the layout of the navigation header itself. And I think they work and look a lot better.

This commit is about making sure that 'Exit' is visible in horizontal nav drawer view. So right now I am adding a dummy menu item - just to fill in the space and make sure that the menu fully scrolls over the bottom ad. 

Of course the better solution would be to have the ad appear under the navigation drawer - so we see the entire scrollable area or make sure, in another way, that the scrolling in horizontal mode revels each menu item - not leaving some behind the ad.

I am not sure how this can be done in our case.

marcelinkaaa:
Fully understand. I could not find a better solution as of now, will keep looking. But would not want to release when one of the menu items is not visible to Basic users.

gubatron:
i don't understand what is the issue, that you have to scroll if on horizontal mode on the phone?

marcelinkaaa:
scroll in the horizontal view and see if 'Exit' option is visible. Its behind the ad on Basic.

aldenml:
Well, I don't see any other way than this solution. I guess it's another missing reasonable feature in the framework.

PR #398, opened 2017-03-18
Fixes broken link on Language change

muckachina:
This is a minor issue when trying to change the Language from the 'flag' icon at the bottom bar, the Language-Window has a link 'Help to translate FrostWire' that was pointing the old frostwire-desktop repo. I didn't open an issue because is a tiny thing.


gubatron:
Thank you!
thanks

PR #399, opened 2017-03-18
[android] made toolbars consistent across the app

marcelinkaaa:
adjusted sizes, updated styles, added same padding to all toolbars, removed unused header xml


android/src/com/frostwire/android/gui/activities/BuyActivity.java
```diff
@@ -79,6 +79,7 @@ private void purchaseProduct(int tagId) {
     @Override
     protected void initToolbar(Toolbar toolbar) {
         toolbar.setTitle(getActionBarTitle());
+        toolbar.setTitleMarginStart(20);
```
aldenml: why? isn't the one in `AbstractActivity` enough?

marcelinkaaa: Actually it might be, let me double check. I was looking at BuyActivity first and left it there when it worked. Thank you for noticing.

aldenml:
I feel a little uncomfortable with the `toolbar.setTitleMarginStart(20);`, that `20` is in pixels, so it does not represent accurately the desired space across devices. Can't you use the title margins styles attributes?

marcelinkaaa:
@aldenml the last commit works very well as well. Would that be ok?

PR #414, opened 2017-03-31
[android] Fixes file opening issues for default storage

gubatron:
It seems that after SDK 24 we cannot pass a file:// URI to another app via Intent data
since all the permissions to interact with such file are a responsability of our app
android now decided we need to have a content provider for the files we send to other intents

This is working so far for our default save location, but it's problematic when the user
changes the default save location to say an external storage device since the FileProvider
documentation says that we must provide the directories before-hand when we declare
the provider in the manifest.

it is my hope that we'll find a way to manipulate such paths programatically somehow. Perhaps
you've read about this already @aldenml

### DIFF
--- a/android/res/xml/provider_paths.xml
+++ b/android/res/xml/provider_paths.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths xmlns:android="http://schemas.android.com/apk/res/android">
+    <external-path name="external_files" path="."/>
+</paths>
--- a/android/res/xml/provider_paths.xml
+++ b/android/res/xml/provider_paths.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <paths xmlns:android="http://schemas.android.com/apk/res/android">
     <external-path name="external_files" path="."/>
+    <root-path name="external_files" path="/storage/" />
 </paths>
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -302,5 +302,15 @@
             android:configChanges="keyboardHidden|orientation|screenSize" />
         <activity android:name="com.mopub.mobileads.RewardedMraidActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"/>
+
+        <provider
+            android:name="android.support.v4.content.FileProvider"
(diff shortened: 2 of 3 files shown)


android/res/xml/provider_paths.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths xmlns:android="http://schemas.android.com/apk/res/android">
+    <external-path name="external_files" path="."/>
+</paths>
```
gubatron: this file is supposed to have a list of file paths our `FileProvider` will use to create URIs.
This is a problem since we let users pick SD Card paths to save after the app has been installed and we can't foresee the future if this information has to be here.

I hope we can hack our way into the `FileProvider`'s paths programmatically when the user changes the default save location.

aldenml:
If I remember correctly, uris with "file://" are permanently forbidden for IPC going forward, you just can't send and intent with that kind of uri inside. Also the default `FileProvider` should not allow you to expose paths that are not private to your app. It could be the case of a faulty implementation, like my Moto G3, that security in the SD card is not enforced at all, but that will fail in the general case.

But to complicate matters, if you granted the permission via the android file browsing to some path to a third party app, then that permission is stored and you continue develop in your app thinking that everything is OK, only to realize later that if you uninstall/reinstall the apps (in this case the third party one), things will fail.

We delayed working in the problem because at the time, the fraction of that SDK was minimal, but maybe now is the time :)

gubatron:
>Also the default FileProvider should not allow you to expose paths that are not private to your app.

That's the part that worries me and I'd like to do more testing.
The problem is not being able to change the list of paths available to the `FileProvider` in run time, the documentation seems clear that this cannot be done, and everybody in stack overflow land says you can't use `FileProvider` on SDCards, which doesn't make sense to me.

What you say makes more sense, just giving your `FileProvider` a path that's only available to your app.

gubatron:
this worked also on the Sony Experia

aldenml:
Sony Experia? What Android version?

gubatron:
Android 7.0 - Android security patch level December 1, 2016

muckachina:
Version 1.9.1-build-352

I can open Videos, Images, docs and Torrent  files saved within the SD card on Android 6, 5 and 4.2

PR #419, opened 2017-04-10
[android] expand horizontal preview view when no ads are displayed

marcelinkaaa:
@gubatron I am expanding the horizontal preview to the width of the screen when no ads on the right side are being loaded, or a user paid for the no-ad version. 
Of course that can be later modified if you set up the logic to fill in our own placement ads. 
Let me know if it works ok for now. Looks good on my devices.


gubatron:
Gracias!

PR #420, opened 2017-04-10
[android] fixes sizing differences in portrait and landscape mode

marcelinkaaa:
Toolbar heigh changes slightly depending on the devices and portrait/horizontal modes. We kept it set at 54dp, which is good but since the height was not adjusting when switched to horizontal mode the toolbar title was not properly aligned with the icons. As shown on the image (nr.1 is what we have, nr.2 is with adjusted height of the toolbar
toolbar_sizing
)


gubatron:
took me a while to see the vertical alignment issue there, awesome one liner fix.

PR #422, opened 2017-04-11
[android] removes status bar on full screen video preview

marcelinkaaa:
@gubatron This removes and restores the status bar in video preview. With status bar visible it was not fully full screen, and it was bugging me a bit :) 

Now while adding it I discovered another bug that was already there. When you enter video preview and tap on the image of full screen icon before the video starts loading (right away) - the full screen layout is a bit odd (the download button is visible, and sometimes even the ad shows on top).


aldenml:
does this fix the bug?

marcelinkaaa:
No just removes status bar from view. I have no idea what is causing the bug.

PR #436, opened 2017-04-28
Search view design update

marcelinkaaa:
Design update of the transfer screen 'add magnet' field, slight update to the view_clearable_edittext.xml and some code cleanup. 

Also - ‘maxLines = 1’, not wroking correctly as of sdk 25. It replaces deprecated ‘singleLine’ but with a current bug its not working correctly withouth ‘inputType’ being added. Being looked at by android developers.

### DIFF
--- a/android/res/layout/fragment_transfers.xml
+++ b/android/res/layout/fragment_transfers.xml
@@ -18,12 +18,32 @@
  */
 -->
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
-              xmlns:frostwire="http://schemas.android.com/apk/res-auto"
-              android:layout_width="match_parent"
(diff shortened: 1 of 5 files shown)


android/res/layout/fragment_transfers.xml
```diff
@@ -18,12 +18,32 @@
...
+        android:background="@color/basic_blue"
+        android:orientation="vertical">
+
+        <View
+            style="@style/divider_line_dark"
+            android:layout_width="match_parent"
+            android:layout_height="1dp" />
+
+        <!-- Dummy item to prevent EditText from receiving focus at startup -->
+        <LinearLayout
```
aldenml: what a trick, I would like to no need this (and the one in the search field too).

PR #437, opened 2017-05-02
[android] ImageViewer fragment

gubatron:


android/res/layout/fragment_image_viewer.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gubatron: license

android/src/com/frostwire/android/gui/views/AbstractActivity.java
```diff
@@ -187,7 +187,7 @@ protected final void setToolbarView(View view, int gravity) {
         }
     }
 
-    protected final void setToolbarView(View view) {
+    public final void setToolbarView(View view) {
```
gubatron: I'll revert this

android/src/com/frostwire/android/gui/views/AbstractListAdapter.java
```diff
@@ -504,13 +504,17 @@ protected void initCheckBox(View view, T item) {
     }
 
     protected void initTouchFeedback(View v, T item) {
+        initTouchFeedback(v, item, viewOnClickListener, viewOnLongClickListener, viewOnKeyListener);
+    }
+
+    protected void initTouchFeedback(View v, T item, OnClickListener clickListener, OnLongClickListener longClickListener, OnKeyListener keyListener) {
```
gubatron: more flexible touch feedback initialization method for those list adapters that might behave differently

android/src/com/frostwire/android/gui/views/AbstractListAdapter.java
```diff
@@ -504,13 +504,17 @@ protected void initCheckBox(View view, T item) {
     }
 
     protected void initTouchFeedback(View v, T item) {
+        initTouchFeedback(v, item, viewOnClickListener, viewOnLongClickListener, viewOnKeyListener);
```
gubatron: keep backwards compatibility

android/src/com/frostwire/android/gui/adapters/menu/RenameFileMenuAction.java
```diff
@@ -56,7 +62,11 @@ protected void onClick(Context context) {
 
     private void showRenameFileDialog() {
         RenameFileMenuActionDialog.renameAction = this;
-        RenameFileMenuActionDialog.newInstance(getFilePath()).show(((Activity) getContext()).getFragmentManager());
+        RenameFileMenuActionDialog renameFileMenuActionDialog = RenameFileMenuActionDialog.newInstance(getFilePath());
```
gubatron: I'll make `RenameFileMenuActionDialog.newInstance()` more consistent.
Right now I'm using a setter on `RenameFileMenuActionDialog` and on `DeleteFileMenuActionDialog` I'm passing the dialog listener as part of the `newInstance` parameters.

marcelinkaaa:
Currently when you are in selection mode and you tap on an image to select it - it opens it instead.

marcelinkaaa:
Still not sure why in this case the icon in the toolbar are not white according to added theme in toolabar_main.xml, but for now adding this line in themes.xml does the trick:
<item name="actionBarTheme">@style/ThemeOverlay.AppCompat.Dark.ActionBar</item>

android/res/layout/fragment_image_viewer.xml
```diff
@@ -0,0 +1,9 @@
...
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent" android:layout_height="match_parent">
+
+    <ImageView
+        android:id="@+id/fragment_image_viewer_image"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+</RelativeLayout>
```
aldenml: and `\n`

android/src/com/frostwire/android/core/FileDescriptor.java
```diff
@@ -82,6 +84,39 @@ public FileDescriptor(int id, String artist, String title, String album, String
         ensureCorrectMimeType(this);
     }
 
+    public FileDescriptor(Bundle bundle) {
+        this(bundle.getInt("id"),
```
aldenml: this is interesting, since none of the fields is `final`, I suppose that for consistency, it should be a `fromBundle` and use that here instead.

android/src/com/frostwire/android/gui/adapters/menu/FileListAdapter.java
```diff
@@ -134,7 +138,18 @@ public View getView(int position, View view, ViewGroup parent) {
 
         try {
             initCheckableGridImageView((RelativeLayout) view, item);
-            initTouchFeedback(view, item);
+            initTouchFeedback(view, item, new OnClickListener() {
```
aldenml: this is starting to not looks good

android/src/com/frostwire/android/gui/adapters/menu/RenameFileMenuAction.java
```diff
@@ -73,7 +83,9 @@ private void renameFile(String newFileName) {
         if (isValidFileName(newFileName)) {
             fd.filePath = Librarian.instance().renameFile(fd, newFileName);
             fd.title = FilenameUtils.getBaseName(newFileName);
-            adapter.notifyDataSetChanged();
+            if (adapter != null) {
```
aldenml: why the adapter can be null, if that's the case, isn't that an error?

android/src/com/frostwire/android/gui/adapters/menu/RenameFileMenuAction.java
```diff
@@ -133,12 +145,14 @@ protected void initComponents(Dialog dlg, Bundle savedInstanceState) {
             noButton.setOnClickListener(new ButtonOnClickListener(this, false));
         }
 
-        public void rename() {
+        public String rename() {
             EditText input = findView(getDialog(), R.id.dialog_default_input_text);
+            String newFileName = new String(input.getText().toString());
```
aldenml: `new String(`?

android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
```diff
@@ -0,0 +1,194 @@
...
+
+    @Override
+    protected void initComponents(View v) {
+        imageView = findView(v, R.id.fragment_image_viewer_image);
+    }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        if (fd != null) {
```
aldenml: is this really necessary? I expect all the components to keep the state

android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
```diff
@@ -0,0 +1,194 @@
...
+            // immediately, thus never really starting. This is why I check that the activity has lived
+            // more than 100 milliseconds, half the blink of an eye.
+            // (it works with 50ms on the android emulator but it could be a longer time on slower devices)
+            //
+            // Another way to do this is the non-standard way, which is getting a hold of the
+            // back button view and attaching a click listener, which is not recommended.
+            //
+            // Real Solution: Figure out why the hell android is creating and destroying the action mode
+            // so quickly when the activity is created.
+            long now = SystemClock.elapsedRealtime();
```
aldenml: yes, I will go for the "real solution" :)

android/src/com/frostwire/android/gui/views/AbstractActivity.java
aldenml: OK!

android/src/com/frostwire/android/gui/views/AbstractListAdapter.java
aldenml: not looking good to me, something is odd here

aldenml:
@marcelinkaaa do you think there is something wrong with the device? our theme/style?

@gubatron did you test all those menu`.onClick` with screen orientation?

marcelinkaaa:
@aldenml so just went over the AbstractActivity again and noticed that we do set up the toolbar as a actionBar and not a standalone toolbar (setSupportActionBar(toolbar);). According to Android, "Styling of standalone Toolbar is done differently to the standard action bar, and is set directly onto the view" - this is what we have done now, but because we currently set it up as actionbar we should set the styling in theme files. 
One curious thing is that the current styling worked for some pages and not for other. 
However, since we are currently using one toolbar, its perfectly fine to leave it as action bar and apply the global styles in themes - which works very well. But if we ever want to use standalone toolbars for different activities, etc. we know what to look for. 

https://android-developers.googleblog.com/2014/10/appcompat-v21-material-design-for-pre.html

aldenml:
interesting, so we need to modify the theme, is that correct?

marcelinkaaa:
yes, however... Just deleted the themes from the toolbar.xml leaving only the ones in theme file (global ones) and the main activity as well as Apollo have different colors. It looks like for them applying the styles directly into the toolbar.xml works better, while the imageViewer, action mode, work when themes are applies in themes file. So will look at how the toolbar is supplied to them and if it differs.

marcelinkaaa:
For now the only thing that works, is to set up the overlay themes in both toolbar.xml and theme file.

android/res/layout/fragment_image_viewer.xml
gubatron: done.

android/res/layout/fragment_image_viewer.xml
gubatron: done.

android/src/com/frostwire/android/core/FileDescriptor.java
gubatron: you mean something like this?

```java
public static FileDescriptor fromBundle(Bundle bundle)
```

I didn't do it like that because my `toBundle()` method is not static, it's simply a way to get a bundle representation from `this`

I could make it something like this and make it more consistent, but not sure if it's better to have less `static` symbols loaded as possible

```
java
public static Bundle toBundle(FileDescriptor fd)
```

aldenml: I mean in reverse, having `fromBundle` not static and call it in the constructor

android/src/com/frostwire/android/gui/adapters/menu/RenameFileMenuAction.java
gubatron: not an error, simply the action is being used in a context where there's no adapter to work with. wonder how to go about it, just trying to reuse the action. Perhaps have the adapter implement some sort of interface that gets notified, still the checking for a listener would exist, but it'd be more uniform.

android/src/com/frostwire/android/gui/adapters/menu/RenameFileMenuAction.java
gubatron: I made a copy to make sure the returned value wouldn't be a weird reference (mem leak or lost reference) to the dialog that's about to be dismissed.

android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
gubatron: I can remove this, when it resumes the `fd` is still there somehow, was just being cautious

gubatron:
will test the onClicks rotated.

gubatron:
@marcelinkaaa fixing the onClick on selection mode, I kept testing with long clicks duh.

android/src/com/frostwire/android/core/FileDescriptor.java
gubatron: I guess this is what you meant, but I'm not sure, as I end up repeating code in the second constructor

https://github.com/frostwire/frostwire/pull/437/commits/551068ab6d5be6f2bf9f52ff64f5bd63b8c6747f#diff-4d8bf0d1f6b9710a2c1c6bb359420232R88

android/src/com/frostwire/android/gui/adapters/menu/RenameFileMenuAction.java
aldenml: this does nothing in terms of reference or memory leak. This constructor just re-pin the internal memory of the original string. There are only a handful of justified cases for which you want to actually copy the chars of another string and all of them include a proper subset.

android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
```diff
@@ -0,0 +1,187 @@
...
+                    new SendFileMenuAction(context, fd).onClick();
+                    break;
+            }
+            return true;
+        }
+
+        @Override
+        public void onDestroyActionMode(ActionMode mode) {
+            this.mode.finish();
+                getActivity().finish();
```
aldenml: Cmd+Shift+F

PR #468, opened 2017-06-07
[android] added custom horizontal layout for Tabs in Search

marcelinkaaa:
preserving vertical space on small devices.
@gubatron should work well now, but that separate horizontal layout for TabLayout was necessary to set the height properly
screen shot 2017-06-06 at 6 41 34 pm

### DIFF
--- a/android/res/layout-land/view_searchinput_tablayout.xml
+++ b/android/res/layout-land/view_searchinput_tablayout.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+/*
+* Created by Angel Leon (@gubatron), Alden Torres (aldenml)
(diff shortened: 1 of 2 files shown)


android/res/layout-land/view_searchinput_tablayout.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="utf-8"?><!--
```
aldenml: \n

android/res/layout-land/view_searchinput_tablayout_custom_tab.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?><!--
```
aldenml: \n

PR #471, opened 2017-06-07
[android] refreshes transfer menu on ‘clearing’ transfer list

marcelinkaaa:
Before the overflow menu icon on the toolbar of the transfer screen would only refresh (after clearing all the transfer list items) upon tapping on the menu, or when going to navigation and back, but not while the user stays on the screen and clears the list. 
Therefore the overflow menu icon would stay visible with empty list. 

Added refresh so that the icon updates as the user deletes/clears the transfer list. 

Huge thanks @gubatron


aldenml:
right, I see

PR #494, opened 2017-06-28
Added updated include exclude tip

marcelinkaaa:
DO NOT Merge yet :)
@gubatron 
Left to do: connect the dismiss button so that the tip does not shows after a user dismissed it.
I will work on colors a bit later, I think they can be improved.

screenshot_20170627-200359
screenshot_20170627-200431


gubatron:
I'll take it from here, enjoy your mid-week break

PR #498, opened 2017-07-04
[android] updated layouts of special notifications

marcelinkaaa:
Updates to the special notifications appearing above our features. 
On smaller screens the height of the notification was not representative of the one set in layout img.1, that was in part due to the use of 9patch drawable as a background. Also text was done as an image. I updated the layouts to use xml only. img. 2. Height is now preserved on all screens.


android/res/drawable/view_less_results_notification_background.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gubatron: license

android/res/layout/view_less_results_notification.xml
```diff
@@ -18,17 +18,39 @@
...
     android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    android:background="@drawable/view_less_results_background_yellow"
-    android:gravity="center_vertical|center_horizontal"
-    android:orientation="horizontal">
-    <ImageView
-        android:id="@+id/view_less_results_notification_image"
+    android:layout_height="wrap_content">
+
+    <LinearLayout
```
gubatron: @aldenml marcelinkaaa says that this nesting was the only way the layout would preserve the height that she wanted.

android/res/values/strings.xml
```diff
@@ -692,4 +692,7 @@
     <string name="or">or</string>
     <string name="include">Include</string>
     <string name="exclude">Exclude</string>
+    <string name="seeing_less_results">Seeing less or no results?</string>
```
gubatron: this is a nice plus

gubatron: CC: @alejandroarturom , and we're not ready yet, more strings are coming

android/res/layout/view_less_results_notification.xml
aldenml: let's see if it is necessary

PR #500, opened 2017-07-06
[android] Move Engine.create() as soon as possible to avoid Runtime Exceptions

gubatron:
I couldn't find any code where MusicPlaybackService was explicitly started by us.
I'd like to try moving this as  soon as possible to minimize the chance for those Runtime Errors in MusicPlaybackService or anywhere else that might need access to the Engine's Thread Pool

### DIFF
--- a/android/src/com/frostwire/android/gui/MainApplication.java
+++ b/android/src/com/frostwire/android/gui/MainApplication.java
@@ -62,6 +62,7 @@ public class MainApplication extends Application {
             installHttpCache();
 
             ConfigurationManager.create(this);
+            Engine.create(this);
 
             Platforms.set(new AndroidPlatform(this));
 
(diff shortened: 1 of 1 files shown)


aldenml:
`MusicPlaybackService` is created/stoped by the player machinery.

I don't like this change without analysis of what could happen to the rest, specifically `EngineService`, since that service maybe needs `NetworkManager` and `Librarian`, and starting the service right before them, I don't know if there is a problem there.

I'm working a a way to make `Engine` a better singleton implementation.

aldenml:
Look at this https://github.com/frostwire/frostwire/commit/b2248f343665bf0225df6f5c23029847ad453ee2

It seems that it's not the only problem after all, we have a serious creation setup problem

aldenml:
I think that with this https://github.com/frostwire/frostwire/commit/be2860a0561742696ef1bbd1f663ed012c418a42

you can close this PR

PR #513, opened 2017-07-13
[android] Hungarian translation updated

alejandroarturom:


android/res/values-hu/strings.xml
```diff
@@ -648,5 +649,18 @@
     <string name="file_information">Fájlinformációk</string>
     <string name="use_btjunkie">Használja BTJunkie</string>
     <string name="add_transfer">Átvitel hozzáadása</string>
-
```
aldenml: 👍

PR #540, opened 2017-07-25
[android] hides selection checkbox when there is no files on the list, in My Files and My Files -> Filter

marcelinkaaa:
Before even when there was 0 files - a user could still enter selection mode in My Files. Also the check when a user is in My Files -> Filter.


android/src/com/frostwire/android/gui/fragments/BrowsePeerFragment.java
```diff
@@ -648,6 +656,10 @@ public void onFilterComplete(int i) {
         }
     }
 
+    private void hideCheckBoxMenuItem() {
+            checkBoxMenuItem.setVisible(false);
```
gubatron: remember to use only 4 spaces next time for indentation.
I'll get rid of this method since it only has a one line and we need to save up on symbols.

PR #548, opened 2017-08-02
PIA change

marcelinkaaa:
I made all the changes to the text and images to make the switch to PIA. Kademlia said we might go back to Express VPN one day so I wanted to leave the strings for Express VPN for easier switch. 

If you know how to make all of it better to use - instead of commenting few things out - do let me know and I will make the change tomorrow morning. 
Or if you just want me to delete all the commented Express VPN related strings etc. I will do it as well.


aldenml:
I have no better alternative to suggest now, LGTM

PR #566, opened 2017-08-14
[android] Hides filterButton when Connection is down or search brings…

marcelinkaaa:
… in no results

Currently the filter button on search page and the filter drawer were active even when there was no Internet connection at all or there was absolutely no search results present after search even with connection present. So the drawer opened with no filters as there was no search results. 

Wanted to hide the filterButton from the toolbar when no search results are displayed.


gubatron:
sorry, can't merge, too expensive making that internet check whenever the pipeline is updated.

This is inline with the other ticket we have.

aldenml:
good observation

gubatron:
Addressed [here](https://github.com/frostwire/frostwire/commit/8f724cc2b75493696845f8d6946effde9579d64d) :)

PR #576, opened 2017-08-28
[android] Transfers > Seeding tab

marcelinkaaa:


android/res/values/strings.xml
```diff
@@ -702,4 +705,7 @@
     <string name="try_changing_applied_filters">Try changing applied filters</string>
     <string name="no_data_check_internet_connection">Network unavailable, please check your Internet connection.</string>
     <string name="use_pixabay">Use Pixabay</string>
+    <string name="transfers_context_seed_all_transfers">Seed all transfers</string>
```
gubatron: this one got messed up after the merge/rebase

android/src/com/frostwire/android/gui/adapters/menu/ResumeDownloadMenuAction.java
```diff
@@ -32,9 +42,10 @@
  * @author gubatron
  * @author aldenml
  */
-public final class ResumeDownloadMenuAction extends MenuAction {
+public final class ResumeDownloadMenuAction extends MenuAction implements AbstractDialog.OnDialogClickListener {
 
     private final BittorrentDownload download;
+    private static final String DLG_TURN_BITTORRENT_BACK_ON = "DLG_TURN_BITTORRENT_BACK_ON";
```
gubatron: let's make this a dummy `Object` and save memory.

gubatron: `private static final Object DLG_TURN_BITTORRENT_BACK_ON = new Object()`

gubatron: ok, since `onDialogClick(String tag)` does take a String, I suppose we can't do that.
However we can just put a single character on the string instead:
`private static final String DLG_TURN_BITTORRENT_BACK_ON = "0";`

android/src/com/frostwire/android/gui/adapters/menu/ResumeDownloadMenuAction.java
```diff
@@ -58,4 +69,55 @@ protected void onClick(Context context) {
...
+
+        if (ConfigurationManager.instance().getBoolean(Constants.PREF_KEY_NETWORK_BITTORRENT_ON_VPN_ONLY) &&
+                !NetworkManager.instance().isTunnelUp()) {
+            if (getContext() instanceof Activity) {
+                UIUtils.showShortMessage(((Activity) getContext()).getWindow().getDecorView().getRootView(), R.string.cannot_start_engine_without_vpn);
+            } else {
+                UIUtils.showShortMessage(getContext(), R.string.cannot_start_engine_without_vpn);
+            }
+        } else {
+            Engine.instance().getThreadPool().execute(new Runnable() {
```
gubatron: I think this will go away.

gubatron: ok, we'll have to do a little refactor since we'll now have some repeated code from `SeedAction.onBittorrentConnect()` which has a Context-leak free runnable.

gubatron: so in theory, this now will be better

android/src/com/frostwire/android/gui/adapters/menu/StopSeedingAction.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * Created by Angel Leon (@gubatron), Alden Torres (aldenml)
```
gubatron: add @marcelinkaaa

android/src/com/frostwire/android/gui/adapters/menu/StopSeedingAction.java
```diff
@@ -0,0 +1,95 @@
...
+            UIUtils.showTransfersOnDownloadStart(getContext());
+        }
+    }
+
+    public StopSeedingAction(Context context, BittorrentDownload download) {
+        this(context, null, download, null);
+    }
+
+    private void stopSeedingEm() {
+        if (!TransferManager.instance().isMobileAndDataSavingsOn()) {
```
gubatron: not sure where this came from, but this if doesn't make sense here.
Since this is an action, we'll just stop everything regardless of network type and data savings settings.

We might be repeating code here from elsewhere

android/src/com/frostwire/android/gui/adapters/menu/StopSeedingAction.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            if (transferToClear != null) {
+                TransferManager.instance().remove(transferToClear);
+            }
+        }
+    }
+
+    private void stopSeedingBTDownload() {
+        btDownload.pause();
```
gubatron: perhaps this pause should be done after the check below

android/src/com/frostwire/android/gui/fragments/TransfersFragment.java
```diff
@@ -480,17 +573,24 @@ private void setupAdapter() {
 
     private List<Transfer> filter(List<Transfer> transfers, TransferStatus status) {
         Iterator<Transfer> it;
+        it = transfers.iterator();
         switch (status) { // replace this filter by a more functional style
             case DOWNLOADING:
-                it = transfers.iterator();
                 while (it.hasNext()) {
-                    if (it.next().isComplete()) {
+                    if (it.next().isComplete() || it.next().isSeeding()) {
```
gubatron: will fix, moving twice.

android/src/com/frostwire/android/gui/transfers/TransferManager.java
```diff
@@ -303,7 +303,9 @@ public boolean remove(Transfer transfer) {
 
     public void pauseTorrents() {
         for (BittorrentDownload d : bittorrentDownloads) {
-            d.pause();
+            if (!d.isSeeding() && d.isDownloading()) {
```
gubatron: will have to rename this method to avoid confusion -> `pauseDownloadingTorrents()`

android/src/com/frostwire/android/gui/transfers/TransferManager.java
```diff
@@ -482,7 +484,34 @@ public void resumeResumableTransfers() {
...
+                }*/
+                }
+            }
+        }
+    }
+
+    public void seedFinishedTransfers() {
+        List<Transfer> transfers = getTransfers();
+
+        if (!isMobileAndDataSavingsOn()) {
```
gubatron: gotta double check this works.

!isMobile == WIFI
Data Savings On == Download only on Wifi.

I think it's better to put a stop condition, or not show the option at all in the UI when the conditions aren't there.

common/src/main/java/com/frostwire/bittorrent/BTDownload.java
```diff
@@ -241,7 +241,7 @@ public int getProgress() {
 
     @Override
     public boolean isComplete() {
-        return getProgress() == 100;
+        return getProgress() == 100 && !isSeeding();
```
gubatron: I think this might lead to a lot of bugs both in android and desktop.
Many places might be considering seeding transfers as complete.

common/src/main/java/com/frostwire/bittorrent/BTDownload.java
```diff
@@ -130,7 +130,7 @@ public boolean isPaused() {
     }
 
     public boolean isSeeding() {
-        return th.isValid() && th.status().isSeeding();
+        return th.isValid() && th.status().isSeeding() && !isPaused();
```
gubatron: and this one might not even be necessary. `isSeeding()` should already return false if it's paused.

Will revert this changes and deal with this extra logic at the UI layer to not break anything.

gubatron:
TO-DO (taking a break)
 * In TransfersFragment I'll redo the logic on what actions to show to avoid redundancy and unnecessary iterations, currently the code has a lot of repetitions and it's based on hard-coded positions on the tabs, that'll change, will be a lot simpler.
 * Go over those pieces of code that ask about `isMobileAndDataSavingsOn()`
 * When screen is rotated the last selected tab is lost
 * When seeing isn't enabled we should have a notification in the seeding tab
 * If seeding is enabled, there are finished transfers and no seeding transfers, let's not show a blank screen on the Seeding tab. Let's invite the user to seed everything without the user having to figure out that they have to go to another tab to start seeding. I believe that if you click on Seeding, it might be related with wanting to seed. The more seeds we have on the network the healthier the network and the better the experience for Bittorrent clients overall.

Issue #585, opened 2017-09-21
[android] ImageViewerFragment Open Menu Action buggy behavior


gubatron:
Currently the ImageViewerFragment "Open" menu action just opens the same image again with FrostWire.
My guess is that this action was meant to open the image with external image viewers.

Let's make it so that it opens the image with third party image viewers installed in the host system.

gubatron:
Right now it uses the standard `OpenMenuAction.onClick(Context context)` logic, which goes:

```
if (fileType == Constants.FILE_TYPE_PICTURES && fd != null) {
            Intent intent = new Intent(context, ImageViewerActivity.class);
            intent.putExtra(ImageViewerFragment.EXTRA_FILE_DESCRIPTOR, fd.toBundle());
            context.startActivity(intent);
        }
```

perhaps there we should check... "are we already on `ImageViewerFragment`"? If so, then let's try an intent that allows us to open the given file with a third party app.

PR #586, opened 2017-09-22
[android] Open Menu Action to open image with external Image Viewer

marcelinkaaa:
Issue #585: @gubatron yes it meant to open an image in the external image viewer in case a user want to easily go and edit an image or further apply some fun filters we of course wont have as a downloader app. 

But it only should leave the app when in fullScreenMode. Hence the added code. 
As of now when we select one image - and are in selection more - we can use the Open Menu Action to view that selected image in full screen mode within FrostWire.

### DIFF
--- a/android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
+++ b/android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
@@ -232,15 +232,11 @@ public final class ImageViewerFragment extends AbstractFragment {
                 case R.id.fragment_my_files_action_mode_menu_seed:
                     new SeedAction(context, fd, null).onClick();
                     break;
-                case R.id.fragment_my_files_action_mode_menu_open:
-                    if (inFullScreenMode) {
-                        new OpenMenuAction(context, fd).onClick();
-                    } else {
+                case R.id.fragment_my_files_action_mode_menu_open:ast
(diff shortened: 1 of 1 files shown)


android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
```diff
@@ -232,7 +233,14 @@ public void onDialogClick(String tag, int which) {
                     new SeedAction(context, fd, null).onClick();
                     break;
                 case R.id.fragment_my_files_action_mode_menu_open:
-                    new OpenMenuAction(context, fd).onClick();
+                    if (inFullScreenMode) {
```
gubatron: this line should never occur

android/src/com/frostwire/android/gui/fragments/ImageViewerFragment.java
```diff
@@ -232,7 +233,14 @@ public void onDialogClick(String tag, int which) {
...
+                        new OpenMenuAction(context, fd).onClick();
+                    } else {
+                        Intent intent = new Intent();
```
gubatron: this is good

marcelinkaaa:
Removed unnecessary if statement. For some reason I thought that ImageViewerFragment is connected with the selection mode that also had the same action menu. 
Thanks

PR #592, opened 2017-09-27
Issue 590 scrollbar fix

### LINKED ISSUE #590: [android/UX] No more accidental scrolls when browsing search results or my files
Not sure why I haven't brought this up before, but I've always found it really annoying when browsing my files that I accidentally touch the scrollbar on the side and I end up all the way down and then I have to scroll up to get back to where I was.

I've never had such issue with other apps and when I checked Google Photos, this is what they do, they have one big button that can be used as the scrollbar. I wonder if there's a simple way to style our scrollbar to not be so tall, and be a little wider and visible, or if you think that we should have the same Google-ish control they use on Google Photos

scrollbar-google-photos

marcelinkaaa:
#590 There is one thing left right after that. Making sure that apollo's scrollbars behave in the same way. We can easily add the scrollbar style to list_base.xml but I am not sure how to approach adding the same OnScrollListener logic to the ApolloFragment as its already has an OnScrollListener. Any ideas @gubatron


gubatron:
I'll fix the warnings. Merging.

PR #597, opened 2017-10-05
[desktop] consider actual port range setting

gubatron:


aldenml:
The range is to pass a `max_retry_port_bind` value to `BTEngine` via context. Right now, there is no such logic to do that.

gubatron:
oh, I thought it was the same range that could be specified on the "Firewall" config pane.

PR #613, opened 2017-10-16
[android] material dialogs update

marcelinkaaa:
Since we changed the min API and are now closer to supporting only material I wanted to update the dialogs, especially its buttons so its consistent across the app - as before for ex. we would use fully material dialogs in preferences (and for social dialog) but custom dialogs (with different buttons elsewhere). 

The issue with custom preference dialog I had before is fixed. 

dialogs


gubatron:
I can help with this, but I want to continue the work on transfers first, more important.

gubatron:
do you still need help with this or can it be merged?

marcelinkaaa:
No help needed unless you see something you would like me to adjust. Works well.

android/src/com/frostwire/android/gui/views/AbstractPreferenceFragment.java
```diff
@@ -106,7 +106,8 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
             onClick(null, DialogInterface.BUTTON_NEGATIVE);
 
             AlertDialog.Builder builder = new AlertDialog.Builder(context)
-                    .setTitle(get("mDialogTitle"))
+                    // Title set directly in a custom dialog
+                    //.setTitle(get("mDialogTitle"))
```
gubatron: if these won't be here, let's delete these comments

android/src/com/frostwire/android/gui/views/preference/CustomSeekBarPreference.java
```diff
@@ -222,6 +225,10 @@ public void onStopTrackingTouch(SeekBar seekBar) {
             mCurrentValueTextView = view.findViewById(R.id.dialog_preference_seekbar_with_checkbox_current_value_textview);
             mUnlimitedCheckbox = view.findViewById(R.id.dialog_preference_seekbar_with_checkbox_unlimited_checkbox);
             Bundle arguments = getArguments();
+
+            mDialogTitleTextView = view.findViewById(R.id.dialog_preference_seekbar_with_checkbox_title);
```
gubatron: 👍

PR #623, opened 2017-11-10
[android] added scroll view to view_search_progress

marcelinkaaa:
On smaller screens and depending on other currently present notifications on the page, the “No results were found” notification would often go outside of the page. Made it scrollable so that the Retry button is always accessible.


gubatron:
I love you

PR #641, opened 2018-01-24
[WIP] Setting up correclty strings to empty fragments

votaguz:
This pull request is intended to add default strings when fragments are empty correctly.

The idea is basically add to ApolloFragment a method called `getEmptyStringId()` that will return the default string id `R.string.empty_music`. 

Then that method will be overriden by subclasses of ApolloFragment such as `RecentFragment` or `LastAddedFragment` for example.

### DIFF
--- a/android/apollo/src/com/andrew/apollo/ui/fragments/RecentFragment.java
+++ b/android/apollo/src/com/andrew/apollo/ui/fragments/RecentFragment.java
@@ -91,7 +91,7 @@ public final class RecentFragment extends ApolloFragment<SongAdapter, Song> {
 
     @Override
     public void onLoadFinished(final Loader<List<Song>> loader, final List<Song> data) {
-        mDefaultFragmentEmptyString = R.string.empty_last_added;
+        mDefaultFragmentEmptyString = R.string.empty_recent;
         super.onLoadFinished(loader, data);
     }
 }
--- a/android/apollo/src/com/andrew/apollo/ui/fragments/RecentFragment.java
+++ b/android/apollo/src/com/andrew/apollo/ui/fragments/RecentFragment.java
@@ -88,4 +88,9 @@ public final class RecentFragment extends ApolloFragment<SongAdapter, Song> {
             threadPool.execute(() -> restartLoader(true));
         }
     }
+
+    @Override
+    public int getEmptyStringId(){
+        return R.string.empty_recent;
+    }
 }
--- a/android/apollo/src/com/andrew/apollo/ui/fragments/RecentFragment.java
+++ b/android/apollo/src/com/andrew/apollo/ui/fragments/RecentFragment.java
@@ -90,7 +90,8 @@ public final class RecentFragment extends ApolloFragment<SongAdapter, Song> {
     }
 
     @Override
-    public int getEmptyStringId(){
-        return R.string.empty_recent;
+    public void onLoadFinished(final Loader<List<Song>> loader, final List<Song> data) {
+        mDefaultFragmentEmptyString = R.string.empty_last_added;
+        super.onLoadFinished(loader, data);
     }
 }
--- a/android/apollo/src/com/andrew/apollo/ui/fragments/profile/ApolloFragment.java
+++ b/android/apollo/src/com/andrew/apollo/ui/fragments/profile/ApolloFragment.java
@@ -450,7 +450,7 @@ public abstract class ApolloFragment<T extends ApolloFragmentAdapter<I>, I>
             // Set the empty text
             final TextView empty = mRootView.findViewById(R.id.empty);
             if (empty != null) {
-                empty.setText(getString(R.string.empty_music));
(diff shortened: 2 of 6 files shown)


votaguz:
@aldenml @gubatron noob question: Currently i'm having an issue with inheritance in this pull request. As you can see I created a method in `ApolloFragment` class called `getEmptyStringId()` and then did override it `RecentFragment` and `LastAddedFragment`.

At this point empty message is being showed correctly in RecentFragment but in `LastAddedFragment` the history is different. It keeps using the parent default empty string ( `R.string.empty_music` instead of `R.string.empty_last_added` ) 

Did i do something wrong ?

aldenml:
I'm not a fan of this abstraction, looks to me like excessive use of OO polymorphism, but before going into that, are you sure the method removed is the right move? the method in the parent seems to do a lot of more

android/apollo/src/com/andrew/apollo/ui/fragments/RecentFragment.java
```diff
@@ -88,4 +88,9 @@ public void onMetaChanged() {
             threadPool.execute(() -> restartLoader(true));
         }
     }
+
+    @Override
+    public int getEmptyStringId(){
+        return R.string.empty_recent;
```
gubatron: no need for this override, does the same as parent

votaguz:
@aldenml, in a previous conversation @gubatron suggested to add an abstract method to be implemented for all the subclass of `ApolloFragment`. But I feel that it will be break the DIY principle. So, which one should be the path to take in this case?  My idea is to provide a better way to display that string.

aldenml:
Think about it, what do you really want, provide an abstraction to customize a text, with all the disadvantages of providing more API to understand, or just tweak the text in the child `onLoadFinished`. That method can be simplified/reduced to just a couple of lines. Also remove the unnecessary `getString`s

votaguz:
@aldenml @gubatron I come up with this solution to the problem. In first place i was modifying the wrong `LastAddedFragment`.
My new proposal is to declare a new field in parent class `ApolloFragment` and override that field in subclasses. If that field isn't overridden then use the default string.
Seems to be a more simpler way to do it, without declare a new method. 
I take this opportunity to ask: Why we have another `LastAddedFragment` under `profiles` directory ?   We really use this ? I was trying to do debugging in that class and debugger never stops in the method that i was testing. 🤔

aldenml:
btw, take this back and forth only if you want to sharpen your criteria

aldenml:
good point, is this a critical point performance wise?

votaguz:
Well we have several fragments that should do the same when they don't have elements, thats my point of reference now, but i'm not sure how critical it can be. 

And now I have to ask, since we are using `findViewById` again, should we check for `null` return values ? I was checking the [docs](https://developer.android.com/reference/android/view/View.html#findViewById(int))  and says that: 

```Finds the first descendant view with the given ID, the view itself if the ID matches getId(), or null if the ID is invalid (< 0) or there is no matching view in the hierarchy.```

**Comment by aldenml** [comment]
yes, if going my route, the null check is necessary

**Comment by votaguz** [comment]
Ok, that's adds a new concern for me ( I'm talking about my n00b perspective ), we will have to repeat that four lines of code in every fragment that need to do that. 
```
        TextView empty = mRootView.findViewById(R.id.empty);
        if (empty != null) {
            empty.setText(R.string.empty_recent);
        }
```
And I feel that maybe we are breaking the DIY principle. Should worry about that at this point ?

**Comment by aldenml** [comment]
Ok, I'm willing to settle with this solution, I'm putting the comments in the code

**Comment by gubatron** [comment]
stupid note:
@votaguz -> DIY = Do it Yourself 😆 
I think you meant "DRY" ;)

File: android/apollo/src/com/andrew/apollo/ui/fragments/profile/ApolloFragment.java [hunk 2]
```diff
@@ -127,6 +127,8 @@
 
     protected ViewGroup mRootView;
 
+    protected Integer mDefaultFragmentEmptyString = null;
```
gubatron: why not an `int` and set it to `-1`?

android/apollo/src/com/andrew/apollo/ui/fragments/profile/ApolloFragment.java
```diff
@@ -450,7 +452,8 @@ public void onLoadFinished(final Loader<List<I>> loader, final List<I> data) {
             // Set the empty text
             final TextView empty = mRootView.findViewById(R.id.empty);
             if (empty != null) {
-                empty.setText(getString(R.string.empty_music));
+                empty.setText((mDefaultFragmentEmptyString != null) ?
```
gubatron: then here `(mDefaultFragmentEmptyString != -1)`

gubatron:
LGTM

gubatron:
will add this to `487`

PR #643, opened 2018-01-24
[desktop] Translation updated: Spanish

alejandroarturom:


desktop/lib/messagebundles/es.po
```diff
@@ -5142,9 +5128,8 @@ msgid "Search tools"
 msgstr "Herramientas de búsqueda"
 
 #: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/search/SearchResultMediator.java:739
-#, fuzzy
 msgid "All Free Downloads"
-msgstr "Descargas gratuitas legales"
+msgstr "Todas las Descargas gratuitas"
```
votaguz: Maybe should be `descargas` instead of `Descargas` ?

gubatron: sounds good like you say @votaguz , but if we want to be faithful to the english version, we want it capitalized as a text, and this is because this text, if I remember correctly appears on a BIG blue background button on the home page :)

I'd just translate this, thinking of the context as:

`Descargas Gratuitas`

votaguz: I pointed on that because checked all the previous texts in the pull request to have some context 😄 . But the way you say, sounds better.

PR #650, opened 2018-01-30
[WIP][desktop] updating source code to be compatible with Java 9

aldenml:


desktop/src/com/limegroup/gnutella/gui/MacEventHandler.java
```diff
@@ -84,16 +85,16 @@ public void openURI(AppEvent.OpenURIEvent openURIEvent) {
             }
         });
 
-        app.addAppEventListener(new AppReOpenedListener() {
+        /*app.addAppEventListener(new AppReOpenedListener() {
```
aldenml: This in particular is giving me a lot of problems

PR #698, opened 2018-04-12
[android] ogury 3.0.x update

gubatron:


android/proguard-rules.pro
```diff
@@ -7,7 +7,6 @@
 -keep class for.** {*;}
 -keep class if.** {*;}
 -keep class int.** {*;}
--keep class io.presage.** {*;}
```
aldenml: why isn't this necessary anymore?

aldenml:
How do we know that their `NetworkChangeReceiver` and `AlarmReceiver` don't suffer from ANRs?

gubatron:
Since yesterday we received just 1 ANR from 1 user which I reported, they
say they already have it fixed.
So far very stable.

PR #700, opened 2018-04-13
[android] Minor: Non-translatable string

alejandroarturom:


gubatron:
Good find!

gubatron:
Now if you translated it in others you might have to delete it from all of them.

Issue #377, opened 2017-03-03
[android] Add Telugu translation (75 million speakers)


gubatron:
Volunteers wanted to translate FrostWire to the Telugu Indian language, preferably native speakers.

This is the file to translate.
https://github.com/frostwire/frostwire/blob/master/android/res/values/strings.xml

alejandroarturom:
On it today.

gubatron:
![love the smell](https://user-images.githubusercontent.com/163977/39678475-c690cf58-514a-11e8-9c88-1b2594152c94.gif)

PR #704, opened 2018-05-06
[android] New Language: Telugu

alejandroarturom:
#377


android/res/values-te/strings.xml
```diff
@@ -0,0 +1,719 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+/*
+ * Created by Angel Leon (@gubatron), Alden Torres (aldenml)
+ * Copyright (c) 2011-2016, FrostWire(R). All rights reserved.
```
aldenml: the header is outdated, but I can take care of it in `master`

PR #706, opened 2018-05-09
Add copy of GNU GPLv3 to the root directory

rogersachan:
image

This pull request makes it so GitHub recognizes that the project is under the GNU GPLv3 license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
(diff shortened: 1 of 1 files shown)


gubatron:
Since this is a mono-repo this might not be a good idea, we might have future projects with other licenses (Apache, MIT, etc)

We already have a GPL [`LICENSE.txt`](https://github.com/frostwire/frostwire/blob/master/desktop/LICENSE.txt) file on the desktop/ project.

The android source code however has no `LICENSE` file in its root, however every source file has a mention of its license in it's headers. We have 657 GNU GPL v3 mentions in source files and 501 Apache v3 mentions in source files

I suppose that if we add a LICENSE file to the android/ folder it should probably contain copies of both licenses in it.

What do you think?

Issue #632, opened 2018-01-09
[desktop] New: Help > Send Feedback


gubatron:
When user clicks on Help > Send Feedback, FrostWire should try to open default Mail client similar to how it occurs on the Android app, and pre-fil the email subject and body with the current FrostWire version, build, OS, and any other relevant details.

If it cannot be done with a default email client, perhaps the feature should integrate with support.frostwire.com (zendesk)

Thanks @muckachina for this suggestion.

muckachina:
Thanks! :)

Tested from git on Mac and Windows and it works properly.

However, @gubatron I've a question:
Where does the feedback go when the user does not include an email address? do you get it somewhere else? If not then the email box should be mandatory no optional.

gubatron:
Yes, you will get it at support@frostwire.com which in turn appears as a ticket in Zendesk, you will just not be able to reply as the email will then come from "noreply@frostwire.com" when no email has been provided by the user. Otherwise you can reply.

gubatron:
Sent you a message and left some explanations. seems to be working fine on this end so far.

gubatron:
ok, not seeing the email, will check if this has anything to do with prime.frostwire.com currently being on a test branch, we're testing integration with tightrope for an offer before downloading frostwire from the website.

gubatron:
@muckachina  the email arrived fine, but I don't see it on zendesk for some reason.

![screen shot 2018-05-22 at 6 55 04 pm](https://user-images.githubusercontent.com/163977/40397760-bbd80c12-5df1-11e8-9536-f7abd970a601.png)

gubatron:
will try again entering a from email

gubatron:
Mistery solved, Zendesk was automatically suspending the emails because it notices they come from a system user :), will see if this behavior can be avoided.

![screen shot 2018-05-22 at 6 59 40 pm](https://user-images.githubusercontent.com/163977/40397893-61053c78-5df2-11e8-834e-8109cfaa9e41.png)

These were under "Suspended Tickets"
https://frostwire.zendesk.com/agent/filters/suspended

muckachina:
I saw them finally :) Thanks!

I created a whitelist on zendesk settings to allow emails from noreply@frostwire.com and now it works properly, all message with or without emails goes to the dashboard.

Also the time lapse is working properly too.

Just one thing, If you don't mind, change the title of the email to something like "Feedback Noreply" or something similar instead of "Got a name though"

PR #707, opened 2018-05-14
[desktop][issue-632] Send Feedback Dialog

votaguz:
Issue #632 
- Includes Send Feedback Dialog 
- Cleanup and Refactors

### DIFF
--- a/desktop/src/com/limegroup/gnutella/gui/GUIMediator.java
+++ b/desktop/src/com/limegroup/gnutella/gui/GUIMediator.java
@@ -863,6 +863,10 @@ public final class GUIMediator {
         new AboutWindow().showDialog();
     }
 
+    public static void showSendFeedbackDialog() {
+        new SendFeedbackDialog().showDialog();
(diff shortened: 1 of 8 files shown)


gubatron:
Still missing:
- An optional "Your Name (Optional)" field, could be on the same row of "Email". This way if we write back to the user we can call her by name.
- Validating the email field if it's not empty, it should be a valid email address. use regex.
- Make sure an user cannot send too many feedback messages too quickly, to do this keep in a new `LongSetting` (declared inside the `SendFeedbackDialog` class properties) the last timestamp of a successfully sent feedback message. Next time the user tries to open the `SendFeedbackDialog` he should get a message box saying "You're doing that too quickly, please wait another <<N> minutes | 1 minute | few seconds> to send more feedback"
-  Validate that a feedback message is at least 16 characters long
- Until all fields are not valid the `Send` button is disabled. The `Cancel` button is only disabled when we hit the `Send` button.

PR #712, opened 2018-05-30
[android] added free trial to a NordVPN button

marcelinkaaa:


android/res/values/strings.xml
```diff
@@ -786,7 +786,7 @@
     <string name="could_not_open_transfer_detail_invalid_infohash">Could not open Transfer Details, invalid InfoHash</string>
     <string name="vpn_suggestion_pia" translatable="false">PIA</string>
     <string name="vpn_suggestion_expressvpn" translatable="false">ExpressVPN</string>
-    <string name="vpn_suggestion_nordvpn" translatable="false">NordVPN</string>
+    <string name="vpn_suggestion_nordvpn">NORDVPN\n\Free 7-day trial</string>
```
aldenml: This extra `\` after `\n`, is that right?

PR #718, opened 2018-06-04
REPEAT_ALL mode by default on playlists

votaguz:
- REPEAT_ALL mode by default on playlists


aldenml:
if this works as intended, I'm fine with it, I just not that into the player workflow at this moment

gubatron:
Merging. It's always a nuissance when you install FrostWire that this is not the default behavior.
You suffer through it when you are for example, driving your car and the music stops at the last song and you have an uncomfortable silence, or when you're running and the playlist ends and you have to find a moment to change that. no more.

Issue #13, opened 2015-11-20
[android-player] Remember randomly played songs.


gubatron:
Currently the player does not remember what song was played before in shuffle mode.
I'd be nice to implement a stack that gets popped to retrieve the last played audio file when the user hits 'previous' during shuffle mode. If user decides to go forward again, then the next song is again a random one.

gubatron:
**There's now a $35 bounty on this issue**

muckachina:
I tested in 2 devices right now build-532 and it seems to be working perfect for Playlists @votaguz :)

I still have to test this release in other devices and double check the behavior on "My Files > My Audios", the Shuffle option could be tricky there with files that are not within the same album/playlist although the back button seems Ok there too. So I'm closing here and I'll re-open it If I find a particular scenario where the "back action" in Shuffle mode fails no problem. 

Thanks! <3

PR #725, opened 2018-06-08
[android ] Remember randomly played songs

votaguz:
Removes the old logic to play in shuffle mode  and includes a new one much more simple. #13

### DIFF
--- a/android/apollo/src/com/andrew/apollo/MusicPlaybackService.java
+++ b/android/apollo/src/com/andrew/apollo/MusicPlaybackService.java
@@ -59,11 +59,7 @@ import com.frostwire.util.Logger;
 import com.frostwire.util.Ref;
 
 import java.lang.ref.WeakReference;
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.LinkedList;
-import java.util.Random;
-import java.util.TreeSet;
 
 import static com.frostwire.android.util.Asyncs.async;
 import static com.frostwire.android.util.RunStrict.runStrict;
@@ -215,21 +211,6 @@ public class MusicPlaybackService extends Service {
      * Moves a list to the last position in the queue
      */
     public static final int LAST = 3;
-
-    /**
-     * Shuffles no songs, turns shuffling off
-     */
-    public static final int SHUFFLE_NONE = 0;
-
-    /**
-     * Shuffles all songs
-     */
-    public static final int SHUFFLE_NORMAL = 1;
-
-    /**
-     * Party shuffle
-     */
-    public static final int SHUFFLE_AUTO = 2;
 
     /**
      * Turns repeat off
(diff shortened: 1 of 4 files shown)


gubatron:
Dedicated to @muckachina, she's been asking for this a long time ago.

android/apollo/src/com/andrew/apollo/MusicPlaybackService.java
```diff
@@ -3671,11 +3358,11 @@ public int getQueuePosition() {
          * {@inheritDoc}
          */
         @Override
-        public int getShuffleMode() {
+        public boolean getShuffleMode() {
```
gubatron: Let's rename this method to: `isShuffleEnabled()`

android/apollo/src/com/andrew/apollo/MusicPlaybackService.java
```diff
@@ -1345,70 +1323,14 @@ public void openFileResult(boolean result) {
      *              otherwise.
      * @return The next position to play.
      */
-    private int getNextPosition(final boolean force) {
+    private int getNextPosition(final boolean force, final boolean shuffleMode) {
```
gubatron: rename parameter `shuffleMode` to `shuffleEnabled` or `shuffleOn` (whatever you prefer)

android/apollo/src/com/andrew/apollo/MusicPlaybackService.java
```diff
@@ -455,16 +434,14 @@
 
     private int mMediaMountedCount = 0;
 
-    private int mShuffleMode = SHUFFLE_NONE;
+    private boolean mShuffleMode = false;
```
gubatron: rename to `mShuffleEnabled` or `mShuffleOn`, no need for "modes", it's a switch.

android/apollo/src/com/andrew/apollo/MusicPlaybackService.java
```diff
@@ -291,7 +275,7 @@
      * track instead of skipping to the previous track when getting the PREVIOUS
      * command
      */
-    private static final long REWIND_INSTEAD_PREVIOUS_THRESHOLD = 3000;
+    private static final long REWIND_INSTEAD_PREVIOUS_THRESHOLD = 10000;
```
gubatron: 10 seconds is a little too much, no? please test this a bit more.
Maybe the right number is in between? 5 seconds?

I suppose it rewinds if the current time is bigger than the number given, otherwise it goes back. 10 seconds seems a bit much. I guess I have to try it with you.

votaguz: That was a little mistake, I changed the value just for testing. It should be 3 seconds.

votaguz:
I have the feeling that we should test a little bit more before do merge to master. Maybe @muckachina can help us with that.

muckachina:
I tested in 2 devices right now build-532 and it seems to be working perfect for Playlists @votaguz :) 

I still have to test this release in other devices and double check the behavior on "My Files > My Audios", the Shuffle option could be tricky there with files that are not within the same album/playlist although the back button seems Ok there too, so you can close this issue here and I'll re-open in case I find an scenario where is failing.

Thanks! <3

Issue #727, opened 2018-06-11
[android] Music Player Queue can't be re-organized with Drag and Drop


gubatron:
[android] Music Player Queue can't be re-organized with Drag and Drop

PR #728, opened 2018-06-11
[android] Songs can be reordered in Queue Fragment

votaguz:
[android] Fix drag and drop issue in queue fragment   #727


android/apollo/src/com/andrew/apollo/ui/fragments/QueueFragment.java
```diff
@@ -54,6 +55,8 @@
 public final class QueueFragment extends ApolloFragment<SongAdapter, Song>
         implements DropListener, RemoveListener, DragScrollProfile {
 
+    private static Logger LOG = Logger.getLogger(QueueFragment.class);
```
gubatron: remove this logger

aldenml:
What happened to my question?

gubatron:
hmm, I think your question didn't come through, sorry.

PR #743, opened 2018-07-02
[android] Mixed translations updated

alejandroarturom:


android/res/values-zh/strings.xml
```diff
@@ -672,4 +672,5 @@
     <string name="vpn_find_out_more">了解更多</string>
     <string name="vpn_suggestion_nordvpn">NORDVPN \n 免费试用7天</string>
     <string name="play_store_out_of_sync">Google Play商店需要同步，请等待几分钟</string>
+    <string name="one_file_shared">已在同一Wi-Fi网络上向FrostWire设备共享一个文件。</string>
```
gubatron: no `n_files_shared`?

aldenml: true, so, are you getting familiar with the asian languages :D 

ping @alejandroarturom

alejandroarturom: @aldenml  @gubatron  See #744

PR #751, opened 2018-07-09
[android] Social icons feedback ui changes

marcelinkaaa:
WIP @gubatron I added slack icon to the about page in settings. But I also made some additional changes - so take a look if you would be ok with them. 

I added 'send feedback' and 'join fw chat' to settings menu, and inside of settings -> about I added slack icon to the social media icons, I placed 'changelog' at the top to be closer to ver and build numbers. I also changed the working of FrostWire gear to 'Support Frostwire' - HOWEVER I am not sure if we can do that based on Google rules?


gubatron:
>I also changed the working of FrostWire gear to 'Support Frostwire'

We can have that but only for PLUS. Def. not allowed by Google Play.

You can use the method that checks if it's PLAY or DEBUG/DEV to show that.

For basic, just show nothing. I'll now proceed to review the changes, perhaps I'll add that IF myself after the merge for you.

android/res/values/strings.xml
```diff
@@ -790,4 +790,6 @@
     <string name="vpn_find_out_more">Find out more</string>
     <string name="frostwire_start_engine_service_security_exception">FrostWire could not start properly due to lack of permissions. Please grant the required permissions and restart the application</string>
     <string name="play_store_out_of_sync">Play Store needs to synchronize, wait a few minutes</string>
+    <string name="join_frostwire_chat">Join FrostWire Chat</string>
```
gubatron: Moar' Android strings coming your way @alejandroarturom.

PR #752, opened 2018-07-10
[desktop] Italian translation updated

alejandroarturom:


desktop/lib/messagebundles/it.po
```diff
@@ -2941,37 +2939,34 @@ msgstr "Usa sempre questa risposta"
...
-#, fuzzy
 msgid "Your Name (Optional)"
-msgstr "Altre opzioni"
+msgstr "Il tuo nome (opzionale)"
 
 #: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/SendFeedbackDialog.java:187
-#, fuzzy
 msgid "Sending"
-msgstr "Distribuzione seme"
+msgstr "Invio"
```
gubatron: Inviando

gubatron: (so the user can have UX feedback when the button is pressed)

gubatron: (I checked google translate, and I understand "Invio" might also be correct, but in this context "Sending" means "Sending out", which can be translated to "Inviando")

PR #768, opened 2018-07-26
[desktop] Dutch translation updated

alejandroarturom:


desktop/lib/messagebundles/nl.po
```diff
@@ -4080,23 +4044,20 @@ msgstr ""
...
-#, fuzzy
 msgid "Bitcoin"
-msgstr "bitcoins"
+msgstr ""
 
 #: /Users/gubatron/workspace.frostwire/frostwire/desktop/src/com/limegroup/gnutella/gui/menu/HelpMenu.java:62
-#, fuzzy
 msgid "Bitcoin Cash"
-msgstr "bitcoins"
+msgstr ""
```
gubatron: will test what happens here, if it puts the original msgid or if it translates it to empty

PR #769, opened 2018-07-27
[desktop] Dutch translation updated 2

alejandroarturom:


alejandroarturom:
@gubatron Now everything should work fine, right? 👯‍♂️

gubatron:
let me create the new messages.jar and we should see.

PR #775, opened 2018-07-31
[android] Thai translation 07312018

alejandroarturom:


gubatron:
Thank you for the translations sir!

PR #784, opened 2018-08-05
[android] Apollo Czech translation Updated

alejandroarturom:


android/apollo/res/values-cs/strings.xml
```diff
@@ -126,4 +126,10 @@
   <string name="empty_search">Nenalezeny žádné výsledky</string>
   <string name="empty_favorite">Zde budou zobrazeny všechny skladby označené jako oblíbené.</string>
   <string name="empty_recent">Zde budou zobrazena všechna alba, která byla poslouchána. Pusťte si nějakou hudbu.</string>
+  <string name="now_listening_to_no_artist_available">#NowPlaying \"<xliff:g id="title">%1$s</xliff:g>\" na #FrostWire</string>
```
alejandroarturom: I was thinking that we better leave this hashtag in English. I have searched for equivalent hashtags in other languages, but I have seen that some also use it in English. What do you think @gubatron ?

alejandroarturom: Porque en español incluso qué sería mejor? #Escuchando? #Ahoraescuchando? #Reproduciendo? #AhoraReproduciendo?

aldenml: I agree that hashtags are hard to get right, since they are intended to a global audience, I think english is acceptable in absence of a better local one

gubatron: en espanol creo que podemos intentar, pero es cierto lo que dice alden, si traduces el hashtag entonces se diluye.

Issue #786, opened 2018-08-07
[all] Whitelisting FrostWire on Trackers


gubatron:
Per [community feedback](https://www.reddit.com/r/torrents/comments/956mtu/frostwire_team_asks_for_feedback/e3qgtxh/), FrostWire isn't white listed among private tracker communities.

1. Start announcing the client under torrent client identifier `FW`
2. Submit a patch to [bep_0020](http://www.bittorrent.org/beps/bep_0020.html) once we've rolled out our torrent client identifier
3. Contact /r/trackers and staffs on IRC to submit FrostWire for testing so they can see it reports upload/download statistics accordingly, adjust whatever settings they might require.

Thanks brickfrog2

aldenml:
Given that we are essentially a `libtorrent` wrapper for the bittorrent part, what practical advantages this give to us? I have no problem with this change, but I would like to know

gubatron:
Thinking about it more, an easier way to put it is, right now I bet you can't use FrostWire on most private trackers and that's unacceptable since we know we have a world class bittorrent client here.
It needs to be whitelisted

popcornenthusiast:
I think FrostWire only will be whitelisted when it offers extensible IP filtering capabilities. So, without #702 and even #466 (cause, to be fair, IP filtering alone does almost nothing), it won't be accepted in private trackers or simply by privacy conscious users.

gubatron:
Update:
First step done. [Advertise the client's user agent and fingerprint](https://github.com/frostwire/frostwire/commit/535c71684bc0ed4fff6ec1f9d327238e168941ae) so tracker operators can identify, evaluate and whitelist.

@popcornenthusiast  Next up, I'll try to take a jab at IP Filtering, and will look again at findspies.

Q: do you know any IRC channels of private trackers on freenode so I can start a conversation to see requirements, testing, getting an account.

gubatron:
thanks for that, I do know Ernesto from TorrentFreak, maybe he has contacts

gubatron:
Update: I've started working on the `IP Filter` tool on [this branch](https://github.com/frostwire/frostwire/commits/ip-filter)

popcornenthusiast:
> Update: I've started working on the IP Filter tool on this branch

Great!

And here to expand on something a [reddit user said](https://www.reddit.com/r/torrents/comments/956mtu/frostwire_team_asks_for_feedback/e3qskqs/) and on IP filtering:

[...]

PR #806, opened 2018-08-22
[all] torrent client fingerprint and user-agent support

gubatron:
In relationship to https://github.com/frostwire/frostwire/issues/786


desktop/src/com/limegroup/gnutella/gui/Initializer.java
```diff
@@ -480,6 +481,8 @@ private void startBittorrentCore() {
 
         ctx.enableDht = SharingSettings.ENABLE_DISTRIBUTED_HASH_TABLE.getValue();
 
+        FrostWireUtils.getFrostWireVersionBuild(ctx.version);
```
aldenml: missing logic here?

PR #808, opened 2018-08-25
[android] Shutdown logic revision

gubatron:
@aldenml @muckachina please take a look at this and test this branch.
The purpose of this branch is to perform a passive shutdown of FrostWire, in the past we tried to force kill all services and the process which is against Android's nature, which will try to revive the process anyway.

It's working as expected, at least what I've tested:

1. That ad networks aren't shutdown and they continue to work, both banners and interstitials. Had to perform a reflection hack on a MoPubInterstitial class to reload interstitials that were destroyed upon exit.

2. HTTP and Image loading works after restart

3. Bittorrent transfers work after restart

4. Previous BitTorrent transfers are restored to their previous stage, however, there's a hiccup on which you will see previous transfers in "Error" state for a second, perhaps you know how to fix this, I was "tirando flechas" on how to restore the bittorrent session. See [this](https://github.com/frostwire/frostwire/commit/bcd10152520032ec2944a9e943dff61520f1f5a4#diff-7f3416d798dcaa50e7aed143a06e0fc4R212) which is the only code that I got it to work with, I bet you'll see the flaw. The issue is that when the torrent engine is stopped the `session` object is destroyed, so I try to recreate a new session from scratch. (That code I linked all occurs on a background thread when you stop the debugger there).

**Still untested: Ad-Removal behavior.**

I think this will also fix issues with Ad-Removal and ads coming back, I've yet to test if ads are still hidden after a restart, they should probably be. But now, we don't kill the PlayStore services on Exit, perhaps that was also causing issues.

Playing with fire, but lots of benefits, like not ever having again the crash that occurred when you tried opening frostwire right after a shutdown and it was still shutting down.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -20,7 +20,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.frostwire.android"
     android:installLocation="auto"
-    android:versionCode="558"
+    android:versionCode="559"
     android:versionName="2.0.8">
     <!-- IMPORTANT!! Ignore these, just use as a reference, now it's handled with gradle -->
     <!-- Plus  android:versionCode="9070xyz" (always commit like this, plus!)-->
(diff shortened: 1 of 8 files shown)


aldenml:
What happened here?

gubatron:
wow

gubatron:
let's revert master master

PR #811, opened 2018-08-27
[android] Shutdown logic fix

gubatron:
@aldenml @muckachina please take a look at this and test this branch.
The purpose of this branch is to perform a passive shutdown of FrostWire, in the past we tried to force kill all services and the process which is against Android's nature, which will try to revive the process anyway.

It's working as expected, at least what I've tested:

    That ad networks aren't shutdown and they continue to work, both banners and interstitials. Had to perform a reflection hack on a MoPubInterstitial class to reload interstitials that were destroyed upon exit.

    HTTP and Image loading works after restart

    Bittorrent transfers work after restart

    Previous BitTorrent transfers are restored to their previous stage, however, there's a hiccup on which you will see previous transfers in "Error" state for a second, perhaps you know how to fix this, I was "tirando flechas" on how to restore the bittorrent session. See this which is the only code that I got it to work with, I bet you'll see the flaw. The issue is that when the torrent engine is stopped the session object is destroyed, so I try to recreate a new session from scratch. (That code I linked all occurs on a background thread when you stop the debugger there).

Still untested: Ad-Removal behavior.

I think this will also fix issues with Ad-Removal and ads coming back, I've yet to test if ads are still hidden after a restart, they should probably be. But now, we don't kill the PlayStore services on Exit, perhaps that was also causing issues.

Playing with fire, but lots of benefits, like not ever having again the crash that occurred when you tried opening frostwire right after a shutdown and it was still shutting down.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -20,7 +20,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.frostwire.android"
     android:installLocation="auto"
-    android:versionCode="558"
+    android:versionCode="559"
     android:versionName="2.0.8">
     <!-- IMPORTANT!! Ignore these, just use as a reference, now it's handled with gradle -->
     <!-- Plus  android:versionCode="9070xyz" (always commit like this, plus!)-->
--- a/android/src/com/frostwire/android/gui/activities/MainActivity.java
+++ b/android/src/com/frostwire/android/gui/activities/MainActivity.java
@@ -77,7 +77,6 @@ import com.frostwire.android.gui.views.MiniPlayerView;
 import com.frostwire.android.gui.views.TimerService;
 import com.frostwire.android.gui.views.TimerSubscription;
 import com.frostwire.android.offers.Offers;
-import com.frostwire.android.util.SystemUtils;
 import com.frostwire.platform.Platforms;
 import com.frostwire.util.Logger;
 import com.frostwire.util.Ref;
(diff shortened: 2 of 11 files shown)


aldenml:
take a look at my comments in the original PR https://github.com/frostwire/frostwire/pull/808

gubatron:
I think they're gone :( (your comments)

After my last commit I no longer see "Error" on transfers in the transfer manager after restoring the session, however, it's taking a lot of time for the transfers to appear, about 30 seconds. You will think that the transfers would not restore if you don't wait long enough.

So I bet there's something I could do to trigger their re-appearance.

muckachina:
@gubatron I got an issue with FW Player I cannot reproduce on current version out, perhaps it's related to ads displayed on Player.

To reproduce:

1) Download partial files (audios) from a torrent (for example a FW feature)
2) Once the download is complete select the file on Transfer to open file
3) Click the option to launch the file with FW Player
4) FW Player appears playing the audio file but when going back to FW menu you can hear the song but the Player has disappeared, you can also minimize or Exit FW and the audio keeps Playing and is not on Notifications

muckachina:
Rendering issue on Transfer list here too, after re-starting FW torrent files disappear from Transfer, they re-appear after starting a new torrent download.

gubatron:
@muckachina if you do a pull on this branch, that issue should be gone.

Now there's just an issue showing newly added transfers.

(I'll delete the comment related to the player as it's not part of this pull request)

muckachina:
I've to test more but with this latest merge prior downloads are loading on Transfer on both FW with ads and without ads... however I got some errors when trying to download files from 3 torrent sources right after installing FW, I re-started the app, got the message about 'location permissions' and it worked fine  so I'll check this carefully tomorrow

PR #815, opened 2018-08-30
[android] Apollo Estonian translation updated

alejandroarturom:


aldenml:
At some point I would like to cleanup unused strings from apollo, that should save us a few symbols in the APK

PR #826, opened 2018-09-20
[android] adding initStorageRelatedRichNotifications in onResume

marcelinkaaa:
@gubatron it looks like it just was not called from onResume, only on initComponents. Adding to onResume did the trick for me.


gubatron:
ha, Alden beat me to it. You asked yourself the right question and saw the fragment life cycle. :)

PR #838, opened 2019-04-20
Update links

trollmad3:
The original what is seeding link is taken down and this replaces it. I found an archive on the wayback machine, copied the text and uploaded it on Pastebin.


gubatron:
great initiative, what'll do with the pastebin article you've pasted is that I'm going to copy it into the Wiki here on our Github repository.

If you can find more of those articles in Way Back machine, please open Issues and paste the links here so I can move them over and update the code base.

trollmad3:
@gubatron will do.

PR #912, opened 2020-07-02
Fixed Torrentz2 age (i.e. creationTime)

HimanshuSharma789:
Fixing torrentz2 search age (i.e. creationTime). Previously torrentz2 do not show the creationTime (age of torrent file) in the app.

- **Fixed the REGEX pattern** to extract the Date (in **Torrentz2SearchPerformer**)
[earlier "\<span\>" tag was concatenated along with the Date] So, i fixed that issue.

- Optimized and fixed the **parsing of creationTime** (in **Torrentz2SearchResult**)

### DIFF
--- a/common/src/main/java/com/frostwire/search/torrentz2/Torrentz2SearchPerformer.java
+++ b/common/src/main/java/com/frostwire/search/torrentz2/Torrentz2SearchPerformer.java
@@ -38,7 +38,7 @@ public class Torrentz2SearchPerformer extends TorrentSearchPerformer {
         //https://torrentz2.eu
         //https://torrentz2.unblockninja.com/
         super("torrentz2.unblockninja.com/", token, keywords, timeout, 1, 0);
-        pattern = Pattern.compile("(?is)<dl><dt><a href='(?<infohash>[a-f0-9]{40})'>(?<filename>.*?)</a>.*?<span class='a'><span title='.*?'>(?<age>.*?)</span><span class='s'>(?<filesize>.*?) (?<unit>[BKMGTPEZY]+)</span> <span class='u'>(?<seeds>\\d+)</span><span class='d'>.*?");
+        pattern = Pattern.compile("(?is)<dl><dt><a href='(?<infohash>[a-f0-9]{40})'>(?<filename>.*?)</a>.*?<span class='a'><span title='.*?'>(?<age>.*?)</span>.*?<span class='s'>(?<filesize>.*?) (?<unit>[BKMGTPEZY]+)</span> <span class='u'>(?<seeds>\\d+)</span><span class='d'>.*?");
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/frostwire/search/torrentz2/Torrentz2SearchResult.java
```diff
@@ -96,38 +96,43 @@ public String getTorrentUrl() {
     }
 
     private long parseCreationTime(String dateString) {
-        long result = 0;
+        long result = System.currentTimeMillis();
         try {
             String[] ds = dateString.split(" ");
-            if (ds[1].contains("hours")) {
+            if (ds[1].contains("hour")) {
```
gubatron, line 102: Awesome fixes

given they keep changing from lowercase to uppercase, perhaps it's best that we do this:
```java
// convert all the date string parts to lowercase
for (int i=0; i < ds.length; i ++) {
  ds[i] = ds[i].toLowerCase();
}
```

Then fix all the ifs below to ask for lower string counterparts

HimanshuSharma789: > ds[1].toLowerCase()

would be enough it think

gubatron: I'm not sure if the java compiler will make an optimization for us, it'd be best if we can make sure those are converted to lower case only once.

gubatron: Also, Strings aren't mutable, if you do `s.toLowerCase()` it returns a new copy of the mutated string, but the original instance remains.

common/src/main/java/com/frostwire/search/torrentz2/Torrentz2SearchPerformer.java
```diff
@@ -38,7 +38,7 @@ public Torrentz2SearchPerformer(long token, String keywords, int timeout) {
         //https://torrentz2.eu
         //https://torrentz2.unblockninja.com/
         super("torrentz2.unblockninja.com/", token, keywords, timeout, 1, 0);
-        pattern = Pattern.compile("(?is)<dl><dt><a href='(?<infohash>[a-f0-9]{40})'>(?<filename>.*?)</a>.*?<span class='a'><span title='.*?'>(?<age>.*?)</span><span class='s'>(?<filesize>.*?) (?<unit>[BKMGTPEZY]+)</span> <span class='u'>(?<seeds>\\d+)</span><span class='d'>.*?");
+        pattern = Pattern.compile("(?is)<dl><dt><a href='(?<infohash>[a-f0-9]{40})'>(?<filename>.*?)</a>.*?<span class='a'><span title='.*?'>(?<age>.*?)</span>.*?<span class='s'>(?<filesize>.*?) (?<unit>[BKMGTPEZY]+)</span> <span class='u'>(?<seeds>\\d+)</span><span class='d'>.*?");
```
gubatron, line 41: Good find!

gubatron:
Congratulations on your first merged Pul Request for FrostWire.

Thank you so much!

PR #915, opened 2020-07-09
[common] Adding Idope.se search support

HimanshuSharma789:
Issue-885
Adding Idope.se support for both android and desktop
Also added IdopeSearchPerformerTest file and string translations

### DIFF
--- a/android/src/com/frostwire/android/gui/SearchEngine.java
+++ b/android/src/com/frostwire/android/gui/SearchEngine.java
@@ -26,6 +26,7 @@ import com.frostwire.search.archiveorg.ArchiveorgSearchPerformer;
 import com.frostwire.search.eztv.EztvSearchPerformer;
 import com.frostwire.search.frostclick.FrostClickSearchPerformer;
 import com.frostwire.search.frostclick.UserAgent;
+import com.frostwire.search.idope.IdopeSearchPerformer;
 import com.frostwire.search.limetorrents.LimeTorrentsSearchPerformer;
 import com.frostwire.search.magnetdl.MagnetDLSearchPerformer;
 import com.frostwire.search.nyaa.NyaaSearchPerformer;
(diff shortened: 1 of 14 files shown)


android/src/com/frostwire/android/gui/SearchEngine.java
```diff
@@ -276,6 +277,14 @@ protected boolean isReady() {
         }
     };
 
+    public static final SearchEngine IDOPE = new SearchEngine("Idope", Constants.PREF_KEY_SEARCH_USE_IDOPE) {
+
+        @Override
+        public SearchPerformer getPerformer(long token, String keywords) {
+            return new IdopeSearchPerformer(token, keywords, DEFAULT_TIMEOUT / 2);
```
gubatron, line 284: why half a timeout?

HimanshuSharma789: The page contain max 10 items. So it should be completed in short time. 🙄

gubatron:
Please write to gubatron@gmail.com so I can send your bounty

HimanshuSharma789:
By the way, 
What is the numCrawls variable in SearchPerformer.
I was not able to get that...!

android/src/com/frostwire/android/gui/SearchEngine.java
gubatron: We need to consider that the app can be used from Smartphones in third world countries with slow connectivity.

gubatron: Also remember that FrostWire isn't connecting to that search engine only, it's also connecting to all the other search engines and it might take a while between a request and the handling of a response.

gubatron:
If I remember correctly this is so that it can crawl multiple pages of search results, although there might be another parameter "pages" for that. can't remember I'd have to look at the parent classes above.

HimanshuSharma789:
Actually i was confused how it works. So i decided to set the value to 1.

android/src/com/frostwire/android/gui/SearchEngine.java
HimanshuSharma789: Hmm. Thats true !

PR #922, opened 2020-08-17
[common] Use continue instead of return if setupSaveDir() returns null

pierrezurek:
When using return, if a file or folder was manually deleted, all
remaining torrents in the loop were skipped.


gubatron:
Thank you! Good find.

PR #938, opened 2020-11-15
Add 1337x torrent icon

Aholicknight:
This icon has been cropped to remove the 1337 part and kept the X part of the logo. The icon was grayscaled to fit the colour scheme of the other icons and cropped to 30x30.


Aholicknight:
My bad, I actually cropped it wrong, it should have been 20x20 and not 30x30. Let me close this pull request and make a new one.

Issue #1042, opened 2025-08-13
Linux: Get rid of helper apps options


generic-pers0n:
# The Issue
Currently, FrostWire relies on default applications set in _Helper Apps_ on Linux. Here, the user can configure their default applications, including the browser to open links in, for example. Unfortunately, on some systems, this may require some configuration if their default browser is different than Firefox or Firefox isn't installed. On my system, for example, I use LibreWolf instead of Firefox, and I would need to configure FrostWire so it properly opens links. Additionally, if I were to change my default browser, I would need to update FrostWire's settings to accomodate that change. This also applies to the other configurable programs too.

Ultimately, FrostWire's handling of these programs could be made better I feel, especially in such a way where a) they work across all platforms, and b) you don't need to rely on platform-specific APIs.

# My Solution
After doing some research, I've learned about `java.awt.Desktop`. I've seen that this class is already used in FrostWire in various places, and I'm wondering if we can expand FrostWire's usage of this class more to reduce platform-specific code and make FrostWire more portable in the process. At least for launching links in the user's preferred browser, we can go ahead and use the `Desktop.browse()` method to open URLs in the user's default browser. On Linux, however, the only catch is that you must have GVFS installed. Despite that, I think that not only should this be pretty simple to do, but also that this should be installed by default on some distributions, though I could be wrong about that. On the other hand, for other files, we may be able to use the `Desktop.open()` method for them, although I haven't experiment beyond URL handling via `Desktop.browse()` first.

Aside from making the experience more consistent on across all platforms, this would also help reduce reliance on platform-specific APIs, namely those wrapped around in libSystemUtilities. My main concern around this library is how FrostWire relies on pre-built binaries checked into Git, but that's something for another issue to discuss.

# End Goal
Ultimately, I believe that using the `Desktop` class further in FrostWire can help reduce disparities across different platforms, particularly when it comes to Linux vs Windows and macOS. I also believe it will help in keeping FrostWire code to be cross-platform and to not have to worry about platform-specific details. I would like to open PRs to complete this work, but I feel it is appropriate we discuss on where we should go with this first before anything.

gubatron:
Great initiative.
I'd start by taking a look at libSystemUtilities and seeing what functionality is no longer needed there and provided (truly in a cross OS manner) by JDK 22 and up, We will probably be including JDK 24 in our next major release.

[...]

PR #1043, opened 2025-08-15
Drop Helper Apps options

generic-pers0n:
As promised in #1042, I would create a PR demonstrating those planned changes. Now, that PR has arrived! 😄

This PR replaces various methods with Java's `Desktop.browse()` and `Desktop.open()` methods, respectively. The previous methods would, on some platforms, rely on native code via the `SystemUtilities` class to open a file or URL via platform-specific functionality. Otherwise, and namely for Linux, it would rely on helper apps that users would need to set in order to open files and URLs in user-preferred applications.

The only caveat I can see here is the requirement that GVFS be installed. This isn't made clear in Java's API reference, but from some good-old web searching I found out that the `Desktop.open()` method requires that GVFS be installed. To help with that, if opening a URL fails on Linux, the user is prompted to install GVFS. Note that I'm not entirely sure about `Desktop.open()`, on the other hand, but I assume that the GVFS requirement also applies to that method as well, although I don't do any prompting.

Note: this PR currently fails to build due to jlibtorrent not being found from its repository or so. I assume this is related to on-going work trying to move away from a precompiled jlibtorrent and instead relying on a remote repository. I certainly applaud this effort, so I accept potentially waiting until that gets accomplished, where afterwards I can rebase this PR if desired.


gubatron:
I understand GVFS is Gnome's. Wonder how this will work for KDE users and what the market share is, which I think might be very popular.

According to Grok:
Screenshot 2025-08-15 at 8 48 57 AM

 "KDE captures around 30-50% preference in enthusiast communities, while GNOME hovers at 10-25%, boosted by defaults in beginner-friendly distros. For global Linux desktops (estimated 4-5% of all desktops in 2025 per Statcounter/Steam), GNOME likely edges out in raw user numbers, but KDE is closing the gap."
 
 I'll try to test the branch with KDE, fingers crossed this works well there too.

gubatron:
If all this works and we can reuse this code also for mac and windows, then we should probably clean up the C++ native code:

Some places to look at:

`Shell.h` https://github.com/frostwire/frostwire/blob/master/desktop/lib/native-src/windows/SystemUtils/Shell.h

`Shell.cpp` https://github.com/frostwire/frostwire/blob/master/desktop/lib/native-src/windows/SystemUtils/Shell.cpp#L95

but we need to make sure we really don't need
`JNIEXPORT void JNICALL Java_org_limewire_util_SystemUtils_openFileNative(JNIEnv *e, jclass c, jstring path);`
and perhaps this one if you're going for URL opening as well.

 `JNIEXPORT void JNICALL Java_org_limewire_util_SystemUtils_openURLNative(JNIEnv *e, jclass c, jstring url);`

gubatron:
Looking good on mac

generic-pers0n:
Great to hear!

Since it appears that only Windows had this native code, and anything that references it at this point has been removed, I'll go ahead and remove them for this PR.

I'll try this PR on a default Ubuntu and Fedora install, respectively, to see how this works.

gubatron:
unless you have something else pending this LGTM to merge, I'll bring it over to the branch I'm currently working on and fix anything missing.

gubatron:
IMG20250816154949

Would love to send you a FrostWire T-shirt.
If you are in the US write me in private with the size and shipping address

gubatron at gmail

generic-pers0n:
> unless you have something else pending this LGTM to merge, I'll bring it over to the branch I'm currently working on and fix anything missing.

Oh great! That's all I have for this PR! If all looks good then you can proceed to merge it! 😄

generic-pers0n:
> IMG20250816154949
> 
> Would love to send you a FrostWire T-shirt.
> If you are in the US write me in private with the size and shipping address
> 
> gubatron at gmail

Ooh neat! I'll be in touch! 😄

Issue #1047, opened 2025-08-27
[desktop/linux] Get rid of precompiled native binaries


generic-pers0n:
In #1042, I mentioned libSystemUtilities and how there are pre-checked binaries into Git. I would like to further elaborate on that in this issue.

# The Issue
Currently, FrostWire relies on several precompiled binaries for Windows, macOS (x86 and ARM), and Linux (for x86_64 only as far as I see). This stems from the dependency on libSystemUtilities, which contains useful utilities that FrostWire uses across Windows and macOS. (Technically, it's also for Linux, but it appears unused at the moment). While this works conveniently, there are multiple issues associated with this.

The first issue to cover here is _transparency_. Unfortunately, it is harder to audit those precompiled binaries and consequently easier to insert a backdoor as it isn't as easy to check these native binaries. As the [xz-utils backdoor situation](https://en.wikipedia.org/wiki/XZ_Utils_backdoor) has shown, binary data can make it easier to slip in a backdoor. Given how libSystemUtilities is shipped as precompiled native binaries, I wonder if it could be even easier.

The second issue to also cover is how this limits FrostWire's potential for working nearly flawless on more platforms besides Windows, macOS, and Linux. While the majority of users will use one of those three platforms, FrostWire could benefit from increased compatibility for multiple different platforms as it would allow more in the open source community to ship FrostWire.

# My Solution
To solve these problems, I propose two solutions: either drop the precompiled binaries and have them built automatically each time FrostWire is built or get rid of the library entirely.

## Solution 1: Keep `libSystemUtilities` and compile it with the rest of FrostWire

This solution keeps the "status quo" of the current FrostWire codebase, where it relies on native code on different platforms to do things. This way, not much has to change a whole lot in the codebase itself. However, the build system would need to change. Unfortunately, my knowledge of Gradle is limited, so while I could figure out a way to pull it off, it would take some time. Furthermore, FrostWire does not have any publicly available methods for packaging, so those would need to be updated.

## Solution 2: Replace `libSystemUtilities` with Java code

This is my preferred solution. I presume that FrostWire's codebase would change quite a bit but not radically. Not only would getting rid of libSystemUtilities be better for transparency, but it would also allow us to get rid of redundant methods and replace them with Java methods that can work without the involvement of a native library.

#1043 has already proven that we can take this method. In fact, you may even consider it the first steps of this solution. Nevertheless, at least from my
[...]

PR #1057, opened 2025-08-31
Prebuilt Binary Removal Part 1: Minimize usage and size of libSystemUtilities

generic-pers0n:
This is the first step in dropping prebuilt binaries as described in #1047. I first start with the easy bits, although I make a bit of assumptions. (If needed, I can make changes before merging this PR).

This PR is primarly composed of two halves: the refactoring bits and the clean up bits. To cover the first half, I refactored a few methods to use their equivalent Java counterparts instead. For example, trash handling is refactored to use Java's `Desktop` class on Windows and macOS. Unfortunately, this does not appear to work on Linux, but since FrostWire didn't support it before, I consider this acceptable as we are preserving functionality rather than dropping it. Regardless, this helps with reducing libSystemUtilities usage.

Covering the second half, this PR also drops libSystemUtilities not only for Linux but also for macOS. Surprisingly, libSystemUtilities for macOS was only a single function, of which went unused. This effectively makes libSystemUtilities Windows-only. Alas, macOS still requires prebuilt binaries other than jlibtorrent's stuff (which I won't be focusing on right now), but we'll cover those bits later on.

This PR is more expansive than I originally planned it to be, but it turns out there was a lot of more low-hanging fruit than I initially realized. Plus, I decided to go all out and just take care of this on one fell swoop. I know the general task of #1047 is going to be a long process, so I figured I'd get this in for whatever next minor release of FrostWire there might be (maybe 6.15 or something). I also ask to please not close #1047 after merging this PR because this is only one part in a broad issue.

PR #1058, opened 2025-09-03
libSystemUtilities minimization rebased

generic-pers0n:
This is basically #1057 but rebased on the `jlibtorrent-2.0-migration` branch.

In addition to the original PR, this rebased one contains an extra fix to get the Linux build to work. Now, Linux builds are compiling and running with jlibtorrent 2.x! If desired, however, I can make a separate PR for that fix if needed. It's only 1 commit though, so you may be fine merging it here. Whatever your preference is, you get it!

The original description goes as follows below.

----------

This is the first step in dropping prebuilt binaries as described in https://github.com/frostwire/frostwire/issues/1047. I first start with the easy bits, although I make a bit of assumptions. (If needed, I can make changes before merging this PR).

This PR is primarly composed of two halves: the refactoring bits and the clean up bits. To cover the first half, I refactored a few methods to use their equivalent Java counterparts instead. For example, trash handling is refactored to use Java's Desktop class on Windows and macOS. Unfortunately, this does not appear to work on Linux, but since FrostWire didn't support it before, I consider this acceptable as we are preserving functionality rather than dropping it. Regardless, this helps with reducing libSystemUtilities usage.

Covering the second half, this PR also drops libSystemUtilities not only for Linux but also for macOS. Surprisingly, libSystemUtilities for macOS was only a single function, of which went unused. This effectively makes libSystemUtilities Windows-only. Alas, macOS still requires prebuilt binaries other than jlibtorrent's stuff (which I won't be focusing on right now), but we'll cover those bits later on.

This PR is more expansive than I originally planned it to be, but it turns out there was a lot of more low-hanging fruit than I initially realized. Plus, I decided to go all out and just take care of this on one fell swoop. I know the general task of https://github.com/frostwire/frostwire/issues/1047 is going to be a long process, so I figured I'd get this in for whatever next minor release of FrostWire there might be (maybe 6.15 or something). I also ask to please not close https://github.com/frostwire/frostwire/issues/1047 after merging this PR because this is only one part in a broad issue.


gubatron:
confirmed that deleting to trash still works on macos.

PR #1063, opened 2025-09-04
Fix Archive.org search provider

generic-pers0n:
I was wondering why FrostWire's Archive.org search provider didn't work, so I decided to investigate this. It turns out the issue was really simple: an invalid argument was being passed to the `indent` parameter. Since I didn't know what valid values this parameter accepted, I decided to remove it. In doing so, searching Archive.org from FrostWire works again! Now FrostWire has one more working search provider! 🎉


gubatron:
Wow, excellent subtle find!
Will test and merge

gubatron:
[Merged!](https://github.com/frostwire/frostwire/commit/a62e4a1801d8231ffaea25b5391ffd1c2a6ee743) 🙏

PR #1083, opened 2025-09-14
[common] Fix GloTorrents search

generic-pers0n:
This PR fixes the GloTorrents search provider. My first thought was using https://glodls.to since that URL worked, as was done in #1079. However, visiting the site in Librewolf, it gave me an HTTPS site unavailable warning, so I guess there was a problem with a redirect or something. When visiting the page, however, there was another official domain: https://gtso.cc. I tried updating the GloTorrents search provider to use that one and it worked!


gubatron:
please rebase, can't switch to this branch because of the changes with telluride
Screenshot 2025-09-15 at 1 35 32 PM

gubatron:
hmm, i wonder why all the commits are there, i feel like if we try to merge there might be a lot of conflicts (for other branches that need to rebase as well)

gubatron:
oh, I think you rebased this against `jlibtorrent-2.0-migration`, it'd be better if we had this on top of master so it can be included in 6.14

generic-pers0n:
> oh, I think you rebased this against `jlibtorrent-2.0-migration`, it'd be better if we had this on top of master so it can be included in 6.14

Oops! I'll fix that ASAP.

I think I might need to double check a few things but at least the domain is correct. At least, I get warnings about the performer not working correctly, but in testing I recall seeing at least a GloTorrent search result.

gubatron:
Cherry picked merged a4b4f4e6b08d4b42eeca9b4c08904c0d92039ebb

PR #1091, opened 2025-09-17
[desktop] Drop libdispatch classes

generic-pers0n:
Having noticed these classes were unused after creating #1090, I decided to make this PR to drop them. This should help clean up the codebase further too.

**Note**: I don't have a Mac to test this on, so I'm relying on the fact that IDEA found no other mention of these classes in any way. If this some _does_ break the Mac build, then we can wait until another time.

PR #1095, opened 2025-09-17
Remove unused sun.awt.X11 classes

generic-pers0n:
This PR is made in the same spirit as #1091. My only concern is that these classes had their unused warnings suppressed, but I have some confidence that these classes aren't likely to be used in the future. Either way, these classes are pretty small, so it should be a very small (and trivial) change (hopefully).


gubatron:
Absolutely not needed, these were necessary to get the Window pointer when we used to have a video player.
image

PR #1097, opened 2025-09-18
[common/desktop] Clean up support for old Windows versions

generic-pers0n:
Since it appears we're working on FrostWire 7, and I've had these changes around for a while, I figured now would be a good time to introduce this PR! :smile:

This PR simply cleans up FrostWire's support for older Windows versions. Since 2020, to the best of my knowledge, FrostWire has only supported Windows 10 or later. Therefore, we can drop a few obsolete fields and methods and refactor other methods that took older versions of Windows into account for support.

I'm not sure how many people are still using FrostWire on Windows 7, but I highly doubt that anyone's using the latest version of FrostWire on Windows 95 :wink:

### DIFF
--- a/desktop/src/com/limegroup/gnutella/util/FrostWireUtils.java
+++ b/desktop/src/com/limegroup/gnutella/util/FrostWireUtils.java
@@ -125,31 +125,13 @@ public final class FrostWireUtils {
     }
 
     public static File getUserMusicFolder() {
-        File musicFile;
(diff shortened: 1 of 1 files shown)


gubatron:
Judging by NEW visitors only, not even 3% of users are on windows 7 and 8.

It's mostly Windows 10 with 52% and Windows 11 with 45%.

Screenshot 2025-09-18 at 7 57 43 AM

PR #1103, opened 2025-09-18
WIP: Add Flatpak build support

generic-pers0n:
This PR aims to address #984 by allowing FrostWire to build and run in a Flatpak environment. These are changes that are needed to get FrostWire to build in Flatpak. They shouldn't be wide-reaching, as we're only doing the bare minimum to get FrostWire just to build.

There are more changes to be expected, such as adding an AppStream metadata file. Once everything's done, I'll mark this PR as ready for review.

[Here's the repository for the Flatpak manifest.](https://github.com/generic-pers0n/frostwire-flatpak)

### DIFF
--- a/desktop/README.md
+++ b/desktop/README.md
@@ -103,6 +103,30 @@ Then run the project with:
 
 On Windows, use `gradlew.bat run` instead.
 
+# Updating Flatpak dependencies
+FrostWire uses the [Flatpak Gradle Generator script](https://github.com/flatpak/flatpak-builder-tools/tree/master/gradle)
+to collect dependency information for its Flatpak. If you add, remove, or otherwise update a
+dependency in Gradle, be sure to update the dependencies too.
+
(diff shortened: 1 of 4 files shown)


generic-pers0n:
Currently, there are a few issues I'm aware of that are causing the build to fail in the Flatpak:

- `flatpak-gradle-generator.py` seems to be extracting some invalid URL because `flatpak-builder` is complaining that https://plugins.gradle.org/m2/me/champeau/jmh/me.champeau.jmh.gradle.plugin/0.7.2/me.champeau.jmh.gradle.plugin-0.7.2.jar results in a 404 error.
- Last I recall, `okio` was causing problems. I believe this was because of a lack of metadata considering this is a Kotlin Multiplatform library.

gubatron:
f550edfc2c3a5b767545c34efbba6921dbaaa3b9 jmh removed

PR #1131, opened 2025-09-22
[desktop] Gradle wrapper script updates

generic-pers0n:
This PR does a few things to FrostWire's Gradle wrapper scripts on desktop:

- Update to Gradle 9.1.0 for Java 25 support
- Remove `desktop/gradlew` and `desktop/gradlew.bat` from `.gitignore`.
- Add the Gradle wrapper script for Windows.

According to the [official Gradle docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html), it appears that using the wrapper script is recommended because you can "ensure a reliable, controlled, and standardized execution of the build." For me, I can see this as it allows us to use a specific version of Gradle regardless of whatever is installed, if Gradle is even installed.


gubatron:
this seems redundant to the previously merged PR
Screenshot 2025-09-22 at 7 53 15 AM

generic-pers0n:
> this seems redundant to the previously merged PR Screenshot 2025-09-22 at 7 53 15 AM

That's because the other PR was based on this branch, and I was hoping that this PR be merged first.

To be fair, we can classify the Gradle wrapper updates as Java 25 updates, so it doesn't matter a whole lot.

PR #1132, opened 2025-09-22
[common/desktop] A few Java 25 updates

generic-pers0n:
This PR makes a few slight changes to remove some deprecated classes:

- Drop FrostWire's custom `ThreadPoolExecutor` class. This appears to be an old remnant that can easily be replaced with Java's own version.
- Remove an unused import of a class deprecated for removal.

This PR depends on #1131, so it is marked as a draft for now as we await action on that PR first.


gubatron:
Thank you for this! very good catch with `ThreadPoolExecutor`.

PR #1144, opened 2025-09-24
[common/desktop] Fix Javadoc errors

generic-pers0n:
This PR gets Javadoc generation working for desktop by fixing hundreds of errors preventing its generation. There are still over 1,000 warnings, but the most important part is that all errors have been fixed so that Gradle's `javadoc` task works now.

I feel that having Javadoc is important as it provides a new outlet to study the FrostWire codebase. In addition, it should make searching for classes and their documentation (if available) much easier than having to look through the source code. This will especially help with learning what classes extend other ones and what classes implement interfaces, and vice versa.

### DIFF
--- a/desktop/src/com/limegroup/gnutella/gui/GUIMediator.java
+++ b/desktop/src/com/limegroup/gnutella/gui/GUIMediator.java
@@ -287,7 +287,7 @@ public final class GUIMediator {
 
     /**
      * Returns a `Point` instance containing the x, y position of the
-     * wrapped ` on the screen.
+     * wrapped `JFrame` on the screen.
      *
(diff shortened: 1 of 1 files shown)


generic-pers0n:
To further add, this will be useful in studying FrostWire's current search architecture. This way, I can easily view methods, interfaces, classes, and see how they interact with each other.

gubatron:
@copilot take a look at this pull request and let me know your thoughts

gubatron:
Thank you so much for this!

generic-pers0n:
> Thank you so much for this!

No problem! I'm always glad to help! :smile:

PR #1172, opened 2025-10-03
WIP: [all] Refactor Archiveorg search performer

generic-pers0n:
This PR refactors the Archiveorg search performer to get rid of two classes when parsing JSON data. I dropped the `ArchiveorgResponseField` class as we never used the `start` field anywhere, and, as we both discovered in #1164, we can just parse the data we need directly into a primitive array of `ArchiveorgItem` objects directly. Consequently, the `ArchiveorgResponse` is also dropped.

I also renamed _almost_ everything with `Archiveorg` to `InternetArchive` as I feel that name flows much better. The only things that haven't been renamed are preference keys for enabling the search provider as this would've broken compatibility with previous FrostWire versions. However, since we are working on new major versions, we can also rename this preference too for consistency if we so desire.

If the renaming is not desired, then we can drop the last commit too. I haven't tested the search performer to ensure it still works either, and hence why this PR is also marked as a draft. This is also my first PR touching the Android codebase, so hopefully I didn't break anything there either.


gubatron:
I don't mind the renaming.
Thank you for doing this.
Less is more.

gubatron:
Please make sure to do the name changes for the SearchEngine implementation on the Android Project, pretty sure this will break the android build when I try to merge

generic-pers0n:
It appears that I broke the search performer in the process. I realized that refactoring these different objects is what broke our JSON parsing. I'll double check to see what went wrong and figure out how to fix it.

PR #1174, opened 2025-10-05
[common] Clean up Knaben search performer

generic-pers0n:
#1070 introduced a new search performer for the Knaben database. While this search performer works, it was unnecessarily complex, bloated, and ultimately, if you will, a _slopfest_ of code that I knew could be cleaned up, and that's what this PR does exactly. Luckily, the search performer still works after all these changes, and with all the work done in this PR, I think this makes for a much better Knaben search performer! :smile:

The first thing I started with was that giant `TorrentData` class. There were _way_ too many fields in that class, mainly to account for Knaben's search API presumably using inconsistent fields. However, I don't suspect this is the case, and even if it was, this was _not_ the way to go with it. Only if it was directly parsing it into the object might it have been reasonable, and while I did not see any of that happening (I likely missed it), I feel there should've been a smarter way to do this. Luckily, after making this change, the performer still worked, in the test queries I used, so we should be clear on that front.

The next thing I went on to was to simplify JSON parsing. Since `TorrentData` is now a much more lean class and contains fields that match object names from Knaben's JSON responses, we can go ahead and directly parse this into an array of `TorrentData` objects. However, we need just one more class for that: `Response`. This simply holds a primitive array of `TorrentData` objects, allowing us to parse the JSON data directly into a `Response` object and later iterating through the various `TorrentData` objects. JSON parsing also now uses `JsonUtils` so we don't have to create our own `Gson` instance.

Finally, after the previous two achievements, I removed unnecessary methods, imports, and error checking. I understand that the checking for HTML was left from debugging the search performer, but since it's working now, I think it's safe to drop the HTML checks. With everything put together, the search performer source file now comes in at only 145 lines instead of 394. It should hopefully be much easier to maintain from now on as it should also be much easier to understand what's going on.


generic-pers0n:
Usually, I'll avoid talking boldly because I prefer being modest on a lot of things. However, I saw an opportunity to fix what I saw was a bit of a maintainability issue, and this PR is the fruit of that :smile:

Anyways, this was a little side step of a refactoring adventure I was planning to take. That adventure involves simplifying the various `SearchResult` implementations to only a few classes (if I possibly can). I'm still experimenting with it, but my ultimate goal (if possible) is to make there be only a `SearchResult` class, or maybe even _record_, for _all_ search performers to use. It may be a bit ambitious, but I think I know how to take the first step in that refactoring journey! :smile:

gubatron:
Great work! Thank you for caring and cleaning up.

generic-pers0n:
> Great work! Thank you for caring and cleaning up.

Glad I could do something to help improve this search performer! :smile:

PR #1218, opened 2025-10-17
[all] Drop LimeTorrents

generic-pers0n:
This PR drops FrostWire's search provider. In attempting to fix FrostWire's LimeTorrents search providers, the only URL I could find was behind Cloudflare. All other URLs I found simply didn't resolve, leaving us with no choice but to drop the search provider.

Despite us having to drop this search provider, we should still have better results than the latest stable release (at least on desktop) since we've already fixed three of them. Already does FrostWire provide much more results thanks to these fixes, and I hope to fix more so there are plenty of more search results to come.

Just a little heads up: BTDigg might be next if we can't use their API or their API doesn't work. Maybe I'll try getting in contact with them to see if we can have access to their API to get a working BTDigg search provider again. If, for any reason, any of this fails, then BTDigg will be my next target.


gubatron (approved):

gubatron:
Thank you so much for the cleanup!

generic-pers0n:
> Thank you so much for the cleanup!

Glad I could help once again! :smile:

PR #1231, opened 2025-11-19
[desktop] Add GitHub Actions workflow

generic-pers0n:
This PR adds a new GitHub Actions workflow that builds FrostWire on Ubuntu. Note that it only builds on Ubuntu each time there's a push or pull request to `master`, but I hope it makes it easier to determine if a pull request builds or not. Additionally, we can have the same thing performed on Windows or macOS with no difference, but I'm okay if this PR is merged before I get to do that.

Not only will this workflow be helpful in testing if whether or not FrostWire builds, but it will also open the door to nightly builds. This will require that we open up our packaging, but I'm already looking at replacing the current packaging scripts used with using `jpackage` via Gradle or something. Until that happens, I hope this helps in testing FrostWire across multiple platforms! :smile:

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -44,6 +44,10 @@ jobs:
     - name: Build with Gradle
       run: cd desktop && ./gradlew build
 
+    - name: "[Linux] Build JAR"
+      if: runner.os != 'Windows' && runner.os != 'macOS'
+      run: ./gradlew jar
+
     # Uncomment the following step to enable publishing builds when a release is made.
     #
     # The USERNAME and TOKEN need to correspond to the credentials environment variables used in


generic-pers0n:
I've now added a 64-bit Windows runner and a macOS Apple Silicon runner. I've also verified that the configuration works in a test branch too.

gubatron:
interesting, will check, i'm currently building frostwire 7 installers.
A lot has changed in the project structure.

generic-pers0n:
I've now added an additional Linux-only step: building a JAR file. This is anticipated for #1233, but the JAR file should still work the same.

> interesting, will check, i'm currently building frostwire 7 installers. A lot has changed in the project structure.

Definitely noted! Despite that, the build instructions appear to remain the same (`cd desktop && ./gradlew build`), so it appears that this is so far so good!

To see this workflow in action, you can view runs on a similar branch here: https://github.com/generic-pers0n/frostwire/actions.

generic-pers0n:
So far, the workflow is succeeding too. If not for any installers or other packages, it's for helping with building and testing! :D

gubatron:
Great addition as always @generic-pers0n

PR #1233, opened 2025-11-24
[desktop] Make a distributable JAR file for Linux

generic-pers0n:
This PR makes a few tweaks to the desktop project's Gradle file that makes the JAR filename a bit more presentable. It properly set's the project's version to `7.0-SNAPSHOT` and doesn't manually override the JAR file name. Thus, the JAR file name is now `frostwire-desktop-7.0-SNAPSHOT` rather than just `frostwire`.

Because jlibtorrent is now handled via Maven, it also appears that Gradle automatically handle's its native dependencies. In building a JAR via `./gradlew jar`, under `lib/x86_64`, we now have `libjlibtorrent-[version]-.so` (replace `[version]` with the proper jlibtorrent version). From testing the JAR file as built on Arch Linux, it runs perfectly.

Regarding issues such as having universal Linux distribution packaging, you may be interested in offering a JAR file as a kind of "universal Linux package". It might not be the best format, but I feel it's a good solution to have in the meantime until we figure out Flatpak or similar.

### DIFF
--- a/desktop/build.gradle
+++ b/desktop/build.gradle
@@ -231,8 +231,6 @@ sourceSets {
 jar {
     duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 
-    archiveFileName = 'frostwire.jar'
-
     exclude('META-INF/*.SF', 'META-INF/*.DSA', 'META-INF/*.RSA', 'META-INF/*.MF', 'META-INF/*.md')
 
     from { configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) } }


gubatron:
let me think about this, I believe there's a lot of reasons why it's simply called "frostwire.jar"

gubatron:
would probably have to maintain every other installer builder script we have for every platform.

gubatron:
maybe we can have this in another gradle task if you'd like the output to be more meticulous

generic-pers0n:
> would probably have to maintain every other installer builder script we have for every platform.

I don't anticipate replacing any installer scripts with this PR, _but_ it could be an option alongside the current Linux packaging. I'm confident that a JAR version for Linux will provide at least some form of a "universal package" for Linux, at least until I can figure out the Flatpak situation. Also, with the JAR, it's just a simple `./gradlew jar`, in which the JAR is output to 

> maybe we can have this in another gradle task if you'd like the output to be more meticulous

Potentially, but I think this ties in to FrostWire's packaging situation in general (including for Windows and macOS). Maybe we should have a separate discussion of that first.

> let me think about this, I believe there's a lot of reasons why it's simply called "frostwire.jar"

I have a (maybe?) wild guess: the Linux packaging scripts currently rely on that filename being `frostwire.jar`. That's just my wild guess, in which case I can understand that.

gubatron:
i think it will break a lot of stuff, a lot depends on that simple "frostwire.jar" being on the classpath, would have to recompile the windows launcher, mac launcher, windows installer, windows uninstaller, mac installer build scripts (all propietary and private)

gubatron:
but I don't oppose on you adding another task that builds or renames the jar however you see fit, or maybe you can do the renaming in some other part of the flatpak script

generic-pers0n:
> i think it will break a lot of stuff, a lot depends on that simple "frostwire.jar" being on the classpath, would have to recompile the windows launcher, mac launcher, windows installer, windows uninstaller, mac installer build scripts (all propietary and private)

Ah, I see.

> but I don't oppose on you adding another task that builds or renames the jar however you see fit, or maybe you can do the renaming in some other part of the flatpak script

I think I can work with the current name as-is. This PR is really just cosmetic changes and nothing that's actually required for anything.

Consequently, I'll be closing this PR. Maybe one day we can return to this, but for now, I prefer function for such a trivial change.

PR #1238, opened 2025-11-30
[desktop] Update JMH plugin to 0.7.3

generic-pers0n:
Just a simple update of the JMH plugin to 0.7.3. I'm not sure why we were using version 0.7.2 before, but I hope that upgrading to 0.7.3 doesn't cause any problems.

### DIFF
--- a/desktop/build.gradle
+++ b/desktop/build.gradle
@@ -12,7 +12,7 @@
 plugins {
     id 'java'
     id 'application'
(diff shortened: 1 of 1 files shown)


gubatron:
I think we can get rid of this JMH plugin, it was used only to run a benchmark, i don't think we'll be doing much of that again.

gubatron:
f550edfc2c3a5b767545c34efbba6921dbaaa3b9 jmh removed

PR #1290, opened 2026-04-21
Close InputStreamReader in readLines

tejas-ae:
While tracing readLines in common/src/main/java/org/apache/commons/io/IOUtils.java I noticed what might be an issue: The resource opened there can leak if readLines exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path. Let me know if the approach doesn’t fit.
Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


tejasae-afk:
Closing — this was a generic resource-leak pattern that isn't the right contribution here.

Issue #622, opened 2017-11-03
[android/UX] Moving completed file to full SD card doesn't show proper error message


gubatron:
Tested on FrostWire Plus 1.9.9-build-463

To reproduce:

Select the SD card as Storage Location (make sure there is few space left on the SD card)

Make a search and start downloading a file from cloud sources (YT, Archive or SC) bigger than the size of the space left within the SD card

The file will go to the Transfer list and will show 'Error moving incomplete' If is a file from YT or SC and it will appear only 'Error' If is a file from Archive.org

It doesn't show the message that indicates 'there is not space left' as it happens with torrent files

—

AzazelSensei:
Opened a PR for the missing disk-full message on cloud/HTTP moves: https://github.com/frostwire/frostwire/pull/1299

PR #1299, opened 2026-08-31
[android] Show disk-full error when a cloud download cannot be moved #622

AzazelSensei:
Cloud/HTTP transfers were showing Error moving incomplete (YouTube/SoundCloud) or a generic Error (Archive.org) when the destination had no space left. Torrents already map that to the no-space-left string.

I pick ERROR_DISK_FULL before notifying if the download fails with No space left on device, and I use the same state when a completed temp file cannot be moved because the destination has less free space than the file.

#622

### DIFF
--- a/android/src/main/java/com/frostwire/android/Android10QFileSystem.java
+++ b/android/src/main/java/com/frostwire/android/Android10QFileSystem.java
@@ -28,6 +28,8 @@ import com.frostwire.android.util.SystemUtils;
 import com.frostwire.platform.DefaultFileSystem;
 
 import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class Android10QFileSystem extends DefaultFileSystem {
     private final Application app;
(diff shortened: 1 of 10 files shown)


gubatron:
Thanks for this submission @AzazelSensei 

The desktop runtime is not affected because desktop uses its own `HttpDownload` implementation. However, this common-code change still needs work for Android:

`moveAndComplete()` infers disk-full from `File.getUsableSpace()`, so permission errors, read-only storage, invalid paths, quotas, and other copy failures can be incorrectly reported as disk-full. It also bypasses FrostWire's `FileSystem` abstraction, which means SAF/`DocumentFile` destinations may be misclassified or not detected at all.

Please propagate the underlying filesystem error or add platform-aware disk-full detection, and add/run tests covering actual copy failures and SAF destinations. The new tests currently are not executed by either the Android or desktop test tasks.

The PR is not malicious and should not break desktop, but its Android fix is incomplete and can provide misleading error states.

AzazelSensei:
Stopped inferring disk-full from File.getUsableSpace. moveAndComplete now uses the FileSystem copy error: only No space left on device is disk-full, permission and SAF failures stay ERROR_MOVING_INCOMPLETE. Desktop test includes com.frostwire.transfers, and I added an Android test for the SAF-style false copy.

gubatron:
I pushed commit `ec7ce8bde` to fix the Android test compilation failure. `Files.readString` is not available in the Android unit-test compilation environment; the test now uses Java 8-compatible stream APIs. Verified with `./gradlew testPlus1DebugUnitTest --tests com.frostwire.transfers.HttpDownloadDiskFullTest` (pass). Next I am addressing per-copy error isolation and Android 10 MediaStore exception propagation.

gubatron:
I pushed commit `c8c2c2e4d` addressing the remaining implementation issues:

- Copy errors are now held in `ThreadLocal` state instead of one shared filesystem field, preventing concurrent downloads from reading each other's errors.
- Android 10+ MediaStore failures now preserve the underlying exception through `Librarian`, so `ENOSPC` can reach the disk-full classifier instead of being replaced by a generic failure.
- Added a regression test for per-thread copy-error isolation.

Verified the focused desktop transfer/platform tests and Android `HttpDownloadDiskFullTest` pass.

gubatron:
Final verification after commits `ec7ce8bde` and `c8c2c2e4d`:

- Desktop focused transfer/platform tests: passed.
- Android `HttpDownloadDiskFullTest`: passed.
- Android unit-test compilation: passed.
- Desktop/common compilation: passed.
- `git diff --check`: passed.
- Concurrent copy-error isolation is covered.
- Android 10+ MediaStore copy exceptions now preserve their underlying cause for classification.

The original Android issue is real and is now addressed without changing desktop's separate HTTP-download implementation. No malicious or suspicious code was found. This is ready to merge.

gubatron:
Thank you so much for your submission.

